PR #3, opened 2017-01-20
Build fix

kevinmoesker:
Small changes to make build.cmd run without issues. First issue was the wrong paket.exe version downloaded via bootstrapper, second an  index.html being removed

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,7 +1,7 @@
 @echo off
 cls
 
-.paket\paket.bootstrapper.exe
+.paket\paket.bootstrapper.exe 4.0.0-alpha045
 if errorlevel 1 (
   exit /b %errorlevel%
 )
--- a/build.fsx
+++ b/build.fsx
@@ -186,7 +186,6 @@ Target "Build" (fun _ ->
 
 Target "BuildClient" (fun _ ->
     run npmTool "install" ""
-    run npmTool "run build" clientPath
 )
 
 
@@ -418,6 +417,7 @@ Target "All" DoNothing
   ==> "AssemblyInfo"
   ==> "Build"
   ==> "BuildClient"
+  ==> "Run"
   ==> "CopyBinaries"
   ==> "GenerateReferenceDocs"
   ==> "GenerateDocs"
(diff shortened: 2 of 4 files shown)


alfonsogarciacaro:
Thanks!

forki:
image


sure this was the right thin?

kevinmoesker:
Hi @forki . You didn't see this error before me touching the code?  I think that the fetch related error is related to the elmish client sample chosen. I saw it but decided not to touch Client side code because I anticipated change on client side code. We could switch the routing elmish example with the simpler counter one and the error would probably not be there. Is that the type of complexity that is suitable for this project? Alternatively this project could show a richer type of client server interaction. @alfonsogarciacaro hinted that possible next steps would be to share DTO's between client and server. What kind of DTO's and what type of miniapplication should be made to demonstrate such sharing?

forki:
No I don't know where the error is coming from. It's not in my original sample

alfonsogarciacaro:
@forki Seems the last version of the commonjs rollup plugin is issuing a lot of warnings, so I've fixed that doing the logical thing: swallowing the warnings if not in `--verbose` mode ;) These warnings are not actual errors so it should be fine.

Anyway, I've removed the whatwg-fetch dependency for now as fetch is already available in most modern browsers. We can add it later to support the old ones... and Safari.

forki:
Another issue is the thing that we now have unstaged files after a build.
I'm not sure what changed.

Am 23.01.2017 09:35 schrieb "Alfonso Garcia-Caro" <notifications@github.com

alfonsogarciacaro:
@forki You mean files in `src/Client/public/bundle`? I removed them, these shouldn't be committed I think.

forki:
yep

2017-01-23 10:30 GMT+01:00 Alfonso Garcia-Caro <notifications@github.com>:

forki:
image

alfonsogarciacaro:
- `src/Client/AssemblyInfo.fs`: The build script shouldn't generate it, as it's useless for the client
- `src/Client/public/bundle`: Weird, [this folder should be ignored](https://github.com/fable-compiler/fable-suave-scaffold/blob/f0903a0c5750eb8c0d405a85b6985a037d75dbbc/src/Client/.gitignore#L18)
- `src/Server/AssemblyInfo.fs`: This is unavoidable if we want Paket to update the AssemblyInfo, it'll constantly update it to latest version.

PR #15, opened 2017-01-22
exec restore based on OS

### LINKED ISSUE #14: Restore on OSX doesn't work
### Description

The project fail to build because of an error in the target.

### Repro steps

Check out with on OSX and try to build.

### Expected behavior

The project should build.

### Actual behavior

It doesn't build. Had 

### Known workarounds

Change this command: https://github.com/fable-compiler/fable-suave-scaffold/blob/master/.paket/Paket.Restore.targets#L10 to use mono on OSX.

mastoj:
Fixes #14.

Haven't verified on Windows sinceI don't have a Windows machine available at the moment, but it did work on OSX and if they haven't changed the OS identifier in targets file it should work.


forki:
I guess it's time for a real dotnet core version of paket.

forki:
And I guess we need the same in paket itself.

mastoj:
Maybe. The build still fails though, I'll try to find time later today to get it working on OS X.

mastoj:
Another error, restore works.

alfonsogarciacaro:
Thanks! I'll try to build on OSX too.

PR #16, opened 2017-01-22
Fix home osx

mastoj:
I'm pretty sure both `../Client/public` works on both *nix and Windows, but `..\Client\public` only works on Windows.


alfonsogarciacaro:
Thanks!

PR #19, opened 2017-01-23
Netcore

forki:
@alfonsogarciacaro could you please take a look at the client build/deploy?


alfonsogarciacaro:
Thanks, I'll run it on my machine and take a look later :+1:

PR #28, opened 2017-02-16
fix intellisense

enricosada:
Pin specific FSharp.NET.Sdk package version

because 1.0.0 is not yet supported by VSCode/Ionide/FSAC

The version range used was `1.0.0-beta-*` but anyway nuget resolved `1.0.0` instead of `1.0.0-beta-060000`


forki:
lol this fixes it.

image

enricosada:
no really. wtf 😭

PR #45, opened 2017-02-21
Reproduce docker root folder issue

forki:
Run the follwoing to see the error:

   build.cmd CreateDockerImage
   docker run -it -p 8085:8085  --rm forki/fable-suave

image

Note: If you revert the Dockerfile to

    FROM microsoft/dotnet:1.1.0-sdk-msbuild-rc4
    COPY /deploy /app
    WORKDIR /app
    EXPOSE 8085
    ENTRYPOINT ["dotnet", "Server.dll"] 

then everything works as expected

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
 FROM microsoft/dotnet:1.1.0-sdk-msbuild-rc4
-COPY /deploy /app
-WORKDIR /app
+COPY /deploy .
 EXPOSE 8085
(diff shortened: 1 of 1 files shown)


forki:
According to @MichaelSimons this is an issue with F# and it works in C#.
So I'm assigning @enricosada to this.

forki:
according to Glenn Condron  it's happening with C# as well.

forki:
Since I assume no one at Microsoft will ever look at it, I'm closing it. Not our bug...

PR #53, opened 2017-02-22
various fixes

dsyme:
Update to JSON.NET latest with F# 4.1 fix

Various other fixes or cosmetic code changes, not all need be taken, I'm still trying to work out what feels most idiomatic

### DIFF
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -158,7 +158,7 @@ let view model dispatch =
 open Elmish.React
 
 // App
-Program.mkProgram init update (lazyView2 view)
+Program.mkProgram init update view
 |> Program.toNavigable hashParser urlUpdate
 |> Program.withConsoleTrace
 |> Program.withReact "elmish-app"
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -133,7 +133,7 @@ let viewPage model dispatch =
     match model.Page with
     | Page.Home ->
         [ words 60 "Welcome!"
-          words 40 "Learn Fabe at http://fable.io" ]
+          a [ Href "http://fable.io" ] [ words 20 "Learn Fable at fable.io" ] ]
 
     | Page.Login -> 
         match model.SubModel with
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -150,7 +150,7 @@ let viewPage model dispatch =
 /// Constructs the view for the application given the model.
 let view model dispatch =
   div []
-    [ id lazyView2 Menu.view model.Menu dispatch
+    [ lazyView2 Menu.view model.Menu dispatch
       hr [] []
       div [ centerStyle "column" ] (viewPage model dispatch)
     ]
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -16,9 +16,9 @@ open Elmish.UrlParser
 // Model
 
 type SubModel =
-| NoSubModel
-| LoginModel of Login.Model
-| WishListModel of WishList.Model
+  | NoSubModel
+  | LoginModel of Login.Model
+  | WishListModel of WishList.Model
 
 type Model =
   { Page : Page
@@ -109,7 +109,7 @@ let update msg model =
     | AppMsg.LoggedIn ->
         let nextPage = Page.WishList
         let m,cmd = urlUpdate (Ok nextPage) model
-        match model.Menu.User with
+        match m.Menu.User with
         | Some user ->
             m, Cmd.batch [cmd; Navigation.modifyUrl (toHash nextPage) ]
         | None ->
@@ -128,10 +128,12 @@ open Fable.Helpers.React
 open Fable.Helpers.React.Props
 open Client.Style
 
+/// Constructs the view for a page given the model and dispatcher.
 let viewPage model dispatch =
     match model.Page with
     | Page.Home ->
-        [ words 60 "Welcome!" ]
+        [ words 60 "Welcome!"
+          words 40 "Learn Fabe at http://fable.io" ]
 
     | Page.Login -> 
         match model.SubModel with
@@ -145,9 +147,10 @@ let viewPage model dispatch =
             [ div [ ] [ lazyView2 WishList.view m dispatch ]]
         | _ -> [ ]
 
+/// Constructs the view for the application given the model.
 let view model dispatch =
   div []
-    [ lazyView2 Menu.view model.Menu dispatch
+    [ id lazyView2 Menu.view model.Menu dispatch
       hr [] []
       div [ centerStyle "column" ] (viewPage model dispatch)
     ]
@@ -155,7 +158,7 @@ let view model dispatch =
 open Elmish.React
 
 // App
-Program.mkProgram init update view
+Program.mkProgram init update (lazyView2 view)
 |> Program.toNavigable hashParser urlUpdate
 |> Program.withConsoleTrace
 |> Program.withReact "elmish-app"
--- a/src/Client/Style.fs
+++ b/src/Client/Style.fs
@@ -18,7 +18,7 @@ let viewLink page description =
       Href (toHash page) ]
     [ unbox description]
 
-let internal centerStyle direction =
+let centerStyle direction =
     Style [ Display "flex"
             FlexDirection direction
             AlignItems "center"
@@ -29,7 +29,6 @@ let internal centerStyle direction =
 let words size message =
     span [ Style [ unbox("fontSize", size |> sprintf "%dpx") ] ] [ unbox message ]
 
-
 let text s = text [] [unbox s ]
 
 let buttonLink cssClass onClick elements = 
@@ -38,10 +37,10 @@ let buttonLink cssClass onClick elements =
         OnTouchStart (fun _ -> onClick())
         Style [ unbox("cursor", "pointer") ] ] elements
 
-let internal onEnter msg dispatch =
-    function
-    | (ev:React.KeyboardEvent) when ev.keyCode = 13. ->
-        ev.preventDefault()
-        dispatch msg
-    | _ -> ()
-    |> OnKeyDown
+let onEnter msg dispatch =
+    OnKeyDown (fun (ev:React.KeyboardEvent) ->
+        match ev with 
+        | _ when ev.keyCode = 13. ->
+            ev.preventDefault()
+            dispatch msg
+        | _ -> ())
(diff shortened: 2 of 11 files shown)


src/Client/App.fs
```diff
@@ -16,9 +16,9 @@ open Elmish.UrlParser
 // Model
 
 type SubModel =
-| NoSubModel
-| LoginModel of Login.Model
-| WishListModel of WishList.Model
+  | NoSubModel
```
dsyme: Various whitespace changes - just whatever felt right to me really

src/Client/App.fs
```diff
@@ -128,10 +128,12 @@ open Fable.Helpers.React
 open Fable.Helpers.React.Props
 open Client.Style
 
+/// Constructs the view for a page given the model and dispatcher.
 let viewPage model dispatch =
     match model.Page with
     | Page.Home ->
-        [ words 60 "Welcome!" ]
+        [ words 60 "Welcome!"
+          words 40 "Learn Fabe at http://fable.io" ]
```
forki: typo ;-)

src/Client/App.fs
```diff
@@ -155,7 +158,7 @@ let view model dispatch =
 open Elmish.React
 
 // App
-Program.mkProgram init update view
+Program.mkProgram init update (lazyView2 view)
```
forki: not sure we should do that. @et1975  ?

src/Client/Style.fs
```diff
@@ -18,7 +18,7 @@ let viewLink page description =
       Href (toHash page) ]
     [ unbox description]
 
-let internal centerStyle direction =
```
dsyme: Given the "light touch" nature of the coding I don't think there's much point in the sample using accessibility controls.

src/Client/Style.fs
```diff
@@ -38,10 +37,10 @@ let buttonLink cssClass onClick elements =
...
 
-let internal onEnter msg dispatch =
-    function
-    | (ev:React.KeyboardEvent) when ev.keyCode = 13. ->
-        ev.preventDefault()
-        dispatch msg
-    | _ -> ()
-    |> OnKeyDown
+let onEnter msg dispatch =
+    OnKeyDown (fun (ev:React.KeyboardEvent) ->
```
dsyme: I don't like the use of "function" and "|>" and so on in this, it's too whack for the beginner.

src/Client/pages/WishList.fs
```diff
@@ -81,7 +82,7 @@ let update (msg:WishListMsg) model : Model*Cmd<WishListMsg> =
     | AddBook ->
         if Validation.verifyBook model.NewBook then
             let wishList = { model.WishList with Books = (model.NewBook :: model.WishList.Books) |> List.sortBy (fun b -> b.Title) }
-            { model with WishList = wishList; NewBook = Book.empty }, postWishListCmd(model.Token,wishList)
+            { model with WishList = wishList }, postWishListCmd(model.Token,wishList)
```
forki: I think this one is actually important. since otherwise we don't clear the input fields

src/Client/pages/WishList.fs
```diff
@@ -81,7 +82,7 @@ let update (msg:WishListMsg) model : Model*Cmd<WishListMsg> =
     | AddBook ->
         if Validation.verifyBook model.NewBook then
             let wishList = { model.WishList with Books = (model.NewBook :: model.WishList.Books) |> List.sortBy (fun b -> b.Title) }
-            { model with WishList = wishList; NewBook = Book.empty }, postWishListCmd(model.Token,wishList)
```
dsyme: Setting the NewBook back to empty while the fields were still filled caused a usability bug e.g. adding multiple books with the same title.

src/Server/WishList.fs
```diff
@@ -34,25 +37,29 @@ let getWishListFromDB userName =
...
 let saveWishListToDB (wishList:WishList) =
     try
         let fi = FileInfo(getJSONFileName wishList.UserName)
         if not fi.Directory.Exists then
             fi.Directory.Create()
         File.WriteAllText(fi.FullName,JsonConvert.SerializeObject wishList)
-    with
-    | _ -> ()
+    with exn ->
+        printfn "Save failed %A" exn
```
forki: this works for debugging. @haf what's the correct way for logging errors in suave?

src/Server/WishList.fs
```diff
@@ -34,25 +37,29 @@ let getWishListFromDB userName =
...
+/// Handle the GET on /api/wishlist
 let getWishList (ctx: HttpContext) =
     Auth.useToken ctx (fun token -> async {
         try
             let wishList = getWishListFromDB token.UserName
             return! Successful.OK (JsonConvert.SerializeObject wishList) ctx
-        with
-        | _ -> return! SERVICE_UNAVAILABLE "Database not available" ctx
+        with exn ->
+            printfn "SERVICE_UNAVAILABLE, %A" exn
```
forki: same here. we should really use the logger framework if we want to keep this

src/Client/pages/WishList.fs
```diff
@@ -13,15 +13,16 @@ open Fable.Core.JsInterop
 open Fable.PowerPack
 open Fable.PowerPack.Fetch.Fetch_types
 
-type Model = {
```
dsyme: Somehow formatting record types with 

```
type foo = {
     ....
}
```
looks really wrong.

src/Client/pages/WishList.fs
```diff
@@ -90,14 +91,14 @@ let update (msg:WishListMsg) model : Model*Cmd<WishListMsg> =
     | FetchError _ -> 
         model, Cmd.none
 
-let addNewBookForm (model:Model) dispatch =
+let newBookForm (model:Model) dispatch =
     let buttonActive = if String.IsNullOrEmpty model.NewBook.Title || String.IsNullOrEmpty model.NewBook.Authors then "btn-disabled" else "btn-primary"
     
-    let titleStatus = if String.IsNullOrEmpty model.NewBook.Title then "" else "has-feedback has-success"
```
dsyme: ``has-feedback`` was used in each case further below anyway where this data is consumed.  Perhaps it should be removed from the places below?

src/Server/WishList.fs
```diff
@@ -69,6 +76,7 @@ let postWishList (ctx: HttpContext) =
                     return! Successful.OK (JsonConvert.SerializeObject wishList) ctx
                 else
                     return! BAD_REQUEST "WishList is not valid" ctx
-        with
-        | _ -> return! SERVICE_UNAVAILABLE "Database not available" ctx
+        with exn -> 
+            printfn "SERVICE_UNAVAILABLE, %A" exn
```
dsyme: Feels like we should always put ``printfn`` or some other logging in catastrophic server-side error cases

src/Client/pages/WishList.fs
forki: I think original reason was it reduces diff if you add fields at beginnging or end ;-) 
Somehow it's used in many places

src/Client/Style.fs
forki: it's probably not even used and go

src/Client/Style.fs
forki: yes that's a good change!

src/Client/pages/WishList.fs
forki: mhm. Tbh I did not test many cases here. usually if some tester in my team is complaining, then I would now model this in a new message and make things explict

src/Client/pages/WishList.fs
forki: that was copied from another sample. I didn't reread it yet, Yes I assume it's ok to remove one

src/Server/WishList.fs
forki: yep the sample should always trace. just not with printfn

src/Client/pages/WishList.fs
et1975: The code was ported from Elm, so yeah... some formatting might be off.

src/Client/pages/WishList.fs
dsyme: OK, but if ``NewBook = Book.empty`` is used (which seems fair enough) then the input fields aren't cleared in the UI (maybe that's the bug).  Well, the authors field is cleared but no others.  This leads dangling text in the fields, and if you press "add book" again it says no text has been addded, see picture.

image

dsyme:
CI fails? 
```
mono  "/home/travis/build/fable-compiler/fable-suave-scaffold/test/UITests/bin/Release/UITests.exe" --fail-on-focused-tests --summary --sequenced --version 
﻿﻿Starting ChromeDriver 2.26.436382 (70eb799287ce4c2208441fc057053a5b07ceabac) on port 46489
Only local connections are allowed.
===[16:30:45 DBG] Initialising BufferManager with 827392
[16:30:45 INF] Smooth! Suave listener started in 194.689 with binding 0.0.0.0:8085
Error: The HTTP request to the remote WebDriver server for URL http://localhost:46489/session timed out after 60 seconds.
Running build failed.
Error:
Expecto test of assembly '/home/travis/build/fable-compiler/fable-suave-scaffold/test/UITests/bin/Release/UITests.exe' failed. Process finished with exit code 255.
Starting FinalTarget: KillProcess
```

src/Server/WishList.fs
dsyme: yes

src/Client/pages/WishList.fs
dsyme: OK fixed here: https://github.com/fable-compiler/fable-suave-scaffold/pull/53/commits/571a85ecb837367de91c98c07941ab1de140dc20

dsyme:
Fixed that problem https://github.com/fable-compiler/fable-suave-scaffold/pull/53/commits/571a85ecb837367de91c98c07941ab1de140dc20

dsyme:
Quite chuffed I was able to fix something :)

dsyme:
@forki I'm done with this for today, heading home.  Thanks, I learned a lot!  Very nice and simple programming model

forki:
regarding CI. Yes we activated UI testing on travis - which was a mistake. I'm trying to fix that together with mighty @lefthandedgoat

src/Client/Style.fs
forki: "and can go out" that is

src/Server/WishList.fs
haf:
```
open Suave.Logging
open Suave.Logging.Message

let logger = Log.create "FableSample"

logger.error (eventX "Save failed with exception" >> addExn exn)
```

For example. It uses the Logary Facade. You could add the Facade to this project, too if you want to.

forki: @dsyme can you please try to use that? thx

PR #54, opened 2017-02-22
offload side-effects into commands

et1975:
Side effects with explicit failure modes are offloaded into commands.

Arguably, logging into console is also a side-effect, but normally you'd have a piece of model dedicated to conveying the problems to the user, so instead of logging into console this would just update the model.

### DIFF
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -75,6 +75,10 @@ let update msg model =
             Page = Page.Login
             SubModel = LoginModel m }, Cmd.batch [cmd; Navigation.modifyUrl (toHash Page.Login) ]
 
+    | StorageFailure e ->
+        console.log("Unable to access local storage:",e)
+        model, []
+
     | LoginMsg msg ->
         match model.SubModel with
         | LoginModel m -> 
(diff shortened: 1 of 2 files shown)


src/Client/App.fs
```diff
@@ -75,6 +75,10 @@ let update msg model =
             Page = Page.Login
             SubModel = LoginModel m }, Cmd.batch [cmd; Navigation.modifyUrl (toHash Page.Login) ]
 
+    | StorageFailure e ->
+        console.log("Unable to access local storage:",e)
```
forki: printfn ?

src/Client/App.fs
```diff
@@ -85,8 +89,8 @@ let update msg model =
                 let newUser : UserData = { UserName = m.Login.UserName; Token = token }
                 let cmd =              
                     if model.Menu.User = Some newUser then cmd else
-                    Utils.save "user" newUser
-                    Cmd.batch [cmd; Cmd.ofMsg LoggedIn ]
+                    Cmd.batch [cmd
+                               Cmd.ofFunc (Utils.save "user") newUser (fun _ -> LoggedIn) StorageFailure ]
```
forki: ah yes!

PR #57, opened 2017-02-23
Use suave logging

forki:
/cc @dsyme 
@haf @Krzysztof-Cieslak where would I create a logger?
I mean I already have one in Server.fs (which comes down in the file list)
Is this a global variable?

### DIFF
--- a/src/Server/WishList.fs
+++ b/src/Server/WishList.fs
@@ -11,6 +11,9 @@ open Suave.RequestErrors
 open System
 open Suave.ServerErrors
 open ServerCode.Domain
+open Suave.Logging
+open Suave.Logging.Message
+
 
 /// The default initial data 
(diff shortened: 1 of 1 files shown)


haf:
The logger itself isn't a global variable, but it does reference one. The one it references is a timestamp-versioned tuple of the configuration and a state.

You can create loggers whereever you want. I normally put them at the top of modules. They're good if you want to get started quickly. All my libraries use them and some minor services. (If you want to go full-out Logary, just use the Facade and it will magically know your types (see the Logary readme))

forki:
ok so https://github.com/fable-compiler/fable-suave-scaffold/pull/57/commits/3999653f24f7fcff00c41889f09aa3a510297b14 would be a good change.
Sorry for being so noobish, but I want to make this a good sample

PR #72, opened 2017-03-16
Canopy cannot enter PW anymore

forki:
Something strange is happening with our UI tests.

wtf

@lefthandedgoat Why can't canopy enter the text anymore!?

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -911,7 +911,7 @@ NUGET
 GROUP Build
 NUGET
   remote: https://www.nuget.org/api/v2
-    FAKE (4.50.1)
+    FAKE (4.55)
 
 GROUP Test
 FRAMEWORK: >= NET45
@@ -921,9 +921,9 @@ NUGET
       FSharp.Core (>= 4.0.1.7-alpha) - framework: >= net463
       NETStandard.Library (>= 1.6) - framework: >= net463
       System.Xml.XDocument (>= 4.0.11) - framework: >= net463
-    canopy (1.1.4)
+    canopy (1.2)
       FSharp.Core (>= 3.0.2)
-      Selenium.WebDriver (3.0)
+      Selenium.WebDriver (>= 3.3)
     Expecto (4.0.3)
       Argu (>= 3.2)
       FSharp.Core (>= 3.1.2.5)
@@ -955,7 +955,7 @@ NUGET
       System.Security.Cryptography.X509Certificates (>= 4.3) - framework: >= net46
       System.Threading.Timer (>= 4.3) - framework: >= net451
     PhantomJS (2.1.1)
-    Selenium.WebDriver (3.0)
+    Selenium.WebDriver (3.3)
     System.AppContext (4.3) - framework: >= net463
     System.Collections.Concurrent (4.3) - framework: >= net463
     System.Console (4.3) - framework: >= net463


forki:
the PR shows that it's red on CI as well

lefthandedgoat:
I am at a client then traveling home today, so I probably can't look at it until tomorrow.

If I clone, switch to canopy branch, what command do I run to repro?  just `sh build.sh`?

Just as a guess, because its all client side stuff, react/angular/whatever is doing some tricky stuff regarding the control and its not a simple input box anymore? or its intercepting the typing and not handling it properly, or the javascript is not done doing its thing before canopy starts typing (most likely)

This may slow it down enough to make it work

After `click "Login"`
Add `on "http://localhost:8085/#login"`

If none of that works I will look at it.

forki:
@lefthandedgoat Weird thing is it worked at some point in time. Now it's broken.
regarding slowness:

    "#username" << "test"

^this should wait for #username to appear right?

that trick with `on` had no effect unfrotunately

forki:
and yes "sh build.sh" should do the run

lefthandedgoat:
Watching the video, it looks like it is waiting for it be there, and its typing before the page is 'done done'.  Because it does register the last few characters of `helllllllllllo`.

Selenium has its metrics on when the page is done loading, and once it is, it lets you type in boxes.  There is a mismatch between when selenium thinks the page is done, and when you can effectively start typing in the box.

Lots of cheesy ways to work around it, I will spend some time coming up with a clean one that works with your page.

alfonsogarciacaro:
@lefthandedgoat Yes, React has the concept of [controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components) for inputs in forms, so this may be causing problems though, as @forki says, it's weird it was working before.

lefthandedgoat:
Working for me on OSX, will try tomorrow on windows VM.

forki:
Yes AFAIK it's also working for @alfonsogcnunez

Am 16.03.2017 20:27 schrieb "Chris Holt" <notifications@github.com>:

lefthandedgoat:
@forki 

I can reproduce it 100% of the time on Windows, but I can not figure out what is wrong.

Things I have tried:

Updated Chrome from 56 to 57
Updated Chrome driver from 2.27 to 2.28
Tried Latest firefox (it fails to write)
Went back to master (older canopy) still fails

There is one error in the javascript window, something about parsing urls.

The other thing I notice is that every key stroke has a logging of "UserMsg" `Updating: Model {Page: Page, Menu: Model, SubModel: SubModel} Navigable {Case: "UserMsg", Fields: Array(1)}`

Every keystroke from selenium is logged like this, but not everything is visible in the box, which makes me think it is not a Selenium problem.

For now I have to give up but I will keep thinking about it.

forki:
@alfonsogcnunez I wonder if we use the wrong  text change event

Am 19.03.2017 02:27 schrieb "Chris Holt" <notifications@github.com>:

alfonsogarciacaro:
@forki Hmm, could be. In the [React docs for controlled components](https://facebook.github.io/react/docs/forms.html#controlled-components) they use the `onChange` event while [fable-suave-scaffold uses `onInput`](https://github.com/fable-compiler/fable-suave-scaffold/blob/eb14b3d5193bde474efe1000ba4f195a3c4deeb1/src/Client/pages/Login.fs#L131). Maybe @et1975 can also help here=

et1975:
Yeah, I had to use OnChange: https://github.com/fable-compiler/fable-elmish/blob/master/samples/react/todomvc/todomvc.fsx#L171

alfonsogarciacaro:
Unfortunately, it seems this commit is still failing :/ https://github.com/fable-compiler/fable-suave-scaffold/commit/15255320b67994b44d4fe926098cd13e0ef636e0

et1975:
I wonder if my `requestAnimationFrame` optimization messed it up... sending a PR.

alfonsogarciacaro:
@et1975 fixed this in #76

PR #75, opened 2017-03-18
Improve install of dotnet core.

rfrerebe:
If an other version of dotnet was already in PATH, first test failed, and full reinstall occured again and again.
Now checking in the temp folder where dotnet core is installed.


alfonsogarciacaro:
Thanks!

PR #76, opened 2017-03-20
PR#72 related, trying rollback to elmish@0.8.2

et1975:


alfonsogarciacaro:
Nice! The build is passing :clap: Maybe, there could be a way to disable `requestAnimationFrame` for the UI tests?

et1975:
Perhaps, but I don't understand the root of the problem... at worst the optimization should only have affected what is rendered, not what the state of the model is. And assuming UI catches up at some point and actually does render, it shouldn't have looked like that (`Value` property should be up to latest, not some right-hand substring).

alfonsogarciacaro:
I guess that's because `requestAnimationFrame` and React are batching the actual changes to the input field and this is confusing Selenium somehow.

If possible I would add a `Program.mkProgramNonOptimized` or similar option without `requestAnimationFrame` (the default would be the optimized one) and call it for test using a compiler directive like `#if TESTS`.

et1975:
I have a theory: if selenium is flooding the input with single characters `requestAnimationFrame` would skip some of the renders, which would cause the OnChange and subsequent model update to be skipped as well.

lefthandedgoat:
Happy its working now.  I am sure that selenium is typing VERY quickly, which could cause problems since things happen in an event loop.

forki:
@et1975 are you going to fix it in elmish?

et1975:
I'm still a bit baffled by what's happening...  even if new UI wasn't rendered the old one still should have handled OnChange. I have a theory, but since I don't have a Windows machine I'll send a PR. If that doesn't work, as @alfonsogarciacaro suggested I could add `withReactUnoptimized` or some such.

PR #79, opened 2017-03-21
trying the version spec revert again

et1975:
Re: pull#78

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,10 @@
 init:
   - git config --global core.autocrlf input
+environment:
+  nodejs_version: "6.0"
+install:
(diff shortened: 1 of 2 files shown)


et1975:
Alright, that seems to do the trick. 
TL;DR: `DefaultValue` has effect only if the TextBox itself is empty, letting automation type as fast as it can and preserving the value.

forki:
So what is the change? Another elmish release? And why are we still red?

et1975:
The change is in the view code, see [pull#78](https://github.com/fable-compiler/fable-suave-scaffold/pull/78). Why travis is failing I don't know, it's unrelated.

Issue #81, opened 2017-03-21
Investigate Travis error 


forki:
For some reason the build can't find fsharp.build.dll on mono. Will take a look on Monday

PR #83, opened 2017-03-22
Create zero diff to investigate #81

forki:


enricosada:
seems ok..

forki:
Weird. It was red yesterday.

alfonsogarciacaro:
I'm no expert in Linux distributions, but it seems in order to use the dotnet SDK we're using the trusty container in Travis, which seems to be less stable with errors that come and go. We had the same problem in Fable some days ago.

forki:
weird

PR #89, opened 2017-04-03
Upgrade to Fable 1.0

alfonsogarciacaro:
To @forki with ❤️


forki:
NICE!

forki:
wow it's green on first shot

forki:
source

It's beautiful!

alfonsogarciacaro:
We have the best project formats, the bestest project formats. It's HUGE! (but smaller at the same time)  

Thanks to @enricosada ;)

enricosada:
👍 

a Q: the `src/Server/App.config` is needed for the ui tests? because the app.config is useless in .net core apps

forki:
mhm it should probably be just removed.

PR #90, opened 2017-04-03
Add Fable.JsonConverter

alfonsogarciacaro:


forki:
It's red

alfonsogarciacaro:
Note to self: be more careful when using VS Code multiple cursors.

forki:
What's the purpose of this change? Does it change encoding?

forki:
Cool.

PR #96, opened 2017-04-11
Remove unused function from bash script

alfonsogarciacaro:
As per @dsyme [comment](https://github.com/fable-compiler/fable-suave-scaffold/commit/a297043dfa72c31a8c631d638a810940afcae8fe#commitcomment-21705496).

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -27,15 +27,6 @@ run() {
   fi
 }
 
-yesno() {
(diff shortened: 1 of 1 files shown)


forki:
where is this even coming from?

alfonsogarciacaro:
¯\_(ツ)_/¯ Probably copy & pasted from an old template.

PR #100, opened 2017-04-16
Fixes client build issues

jeremyabbott:
Several HTML elements were getting passed extra empty lists that were causing the client project to fail building.

** Someone else may already be addressing this, but I fixed it for my local development and thought it might be helpful if someone isn't working on it.  I can create an issue for this if that's more appropriate.**


forki:
Thx

PR #101, opened 2017-04-20
Fix #98: Add InstallServer build target

### LINKED ISSUE #98: Code highlight issue in Server project
Hi guys! And thanks for an absolutely fantastic project!

Having an annoying issue when running it in Visual Studio Code.
Only applies to the Server project.
It all builds and runs fine but the editor itself doesn't recognise (highlights opens with red) namespaces imported with paket in Server.fs and Program.fs but doesn't complain and works well in Shared/Domain.fs.
No issues with Client project either.

Is it something wrong with my setup or a common issue?

Tried on both Mac and Win.

Many thanks 
Yuriy
image

alfonsogarciacaro:
After a change I did to the FAKE script, the `Run` target wasn't running `dotnet restore` on the Server project.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -90,8 +90,11 @@ Target "InstallDotNetCore" (fun _ ->
 // --------------------------------------------------------------------------------------
 // Build library & test project
 
-Target "BuildServer" (fun _ ->
+Target "InstallServer" (fun _ ->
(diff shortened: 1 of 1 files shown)


mizzle-mo:
@alfonsogarciacaro Confirmed working on my mac, building install-server branch.

PR #108, opened 2017-05-14
WishList: reset the three inputs after adding a book

0x53A:
fixes https://github.com/fable-compiler/fable-suave-scaffold/issues/107

Was there a specific reason to use DefaultValue instead of Value?

A different fix would be to keep the previous Book as the current NewBook, that way the Dom <> Model would also stay consistent.

### DIFF
--- a/src/Client/pages/WishList.fs
+++ b/src/Client/pages/WishList.fs
@@ -16,7 +16,8 @@ open Fable.PowerPack.Fetch.Fetch_types
 type Model = 
   { WishList : WishList
     Token : string
-    NewBook: Book
+    NewBook : Book
+    NewBookId : Guid // unique key to reset the vdom-elements, see https://github.com/fable-compiler/fable-suave-scaffold/issues/107#issuecomment-301312224
     TitleErrorText : string option
     AuthorsErrorText : string option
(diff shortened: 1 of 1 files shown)


alfonsogarciacaro:
@0x53A I don't remember exactly the problem but it seems there was indeed a reason to use `DefaultValue`: https://github.com/fable-compiler/fable-suave-scaffold/issues/99#issuecomment-294352110

Maybe we should try the alternative solution instead?

forki:
Maybe both are needed

Am 14.05.2017 10:21 vorm. schrieb "Alfonso Garcia-Caro" <
notifications@github.com>:

alfonsogarciacaro:
Pinging @et1975

et1975:
commented in the issue, TL;DR: pick your poison.

forki:
Is this ready?

0x53A:
I think, yes.

I had changed it so that it keyes the id of the inputs off of a random guid, so that it replaces the inputs in the dom instead of reusing them.

I tested that it
* clears the fields
* doesnt reset the cursor to the end when typing somewhere in the middle.

forki:
Thx

PR #111, opened 2017-05-20
Fix #110: Use Fable with Paket integration

### LINKED ISSUE #110: Build breaks after there recent Fable migration to paket/yarn from npm on Mac
### Description

Build breaks after there recent Fable migration to paket/yarn from npm on Mac 

### Repro steps

./build.sh run 
fails with:

> dotnet fable npm-run start
> ﻿dotnet watch run
> ﻿No executable found matching command "dotnet-watch"
> ERROR: Cannot find Fable.Compiler package. Please make sure it's been added to `paket.dependencies`.

I've updated my Client.proj and Server.proj according to the latest examples but it didn't help very much.
Looks like something is wrong with the build command

alfonsogarciacaro:


paket.dependencies
```diff
@@ -1,14 +1,20 @@
...
 nuget Microsoft.Extensions.Logging.Console
 nuget System.Net.NetworkInformation
 nuget Newtonsoft.Json prerelease
 nuget jose-jwt
 
+# Keep this version synced with dotnet-fable in .fsproj
+nuget Fable.Core 1.0.4
+
+nuget Fable.Elmish.Browser
+nuget Fable.Elmish.Debugger
```
forki: This is crazy cool!

paket.dependencies
```diff
@@ -1,14 +1,20 @@
...
-nuget Fable.Core
 nuget Fable.JsonConverter prerelease
 nuget Suave
 nuget Microsoft.Extensions.Logging.Console
 nuget System.Net.NetworkInformation
 nuget Newtonsoft.Json prerelease
 nuget jose-jwt
 
+# Keep this version synced with dotnet-fable in .fsproj
+nuget Fable.Core 1.0.4
```
forki: Urgs? Why the concrete version?

alfonsogarciacaro: I thought the comment was self-explanatory ;) How else can you be sure that `paket install` will install the same version specified for dotnet-fable in the .fsproj?

alfonsogarciacaro:
Argh, how can I make this work on Travis?

paket.dependencies
forki: Actually no I don't understand. Which versions need to be synchronized?

forki: Ah the dotnetcli reference. Ok now I see.

PR #118, opened 2017-06-18
add gcp deploy example

mastoj:
This PR contains:

* Update to Dockerfile, 1.1.1 --> 1.1.2 (needs to be verified on Azure)
* Update to `Readme` on how to deploy to GCP
* Update to app and test to allow for different port based on env variable
* Configuration file that is used to deploy to GCP


forki:
Awesome

mastoj:
If you try it out be a little bit pasient. The deploy time to gcloud with a custom flex image, which this is, takes some time at the moment. Not sure why, but I really hope they can decrease it some time in the future.

PR #120, opened 2017-06-19
Isolate user login test from changes in wish list.

bentayloruk:
As per commit, *If you remove Isaac's book from the wish list, the Canopy user login test fails, because it checks for Isaac.  While I agree that removing Isaac is a crime, the test should probably check for the logout option in the menu.  Changed to do this.*


forki:
thx

PR #123, opened 2017-06-20
Update to Fable 1.1.3

alfonsogarciacaro:
Updates to Fable _stablish_ 1.1.3, many `prerelease` tags have been removed from paket.dependencies ;) All the dependencies like Fable.Elmish, Fable.React have been also updated with the new Fable.Import.Browser split from Fable.Core.

`build Run` works locally for me, just `build` breaks because it says it cannot find the Expecto assembly, don't know why. Let's see what the CI servers say.


alfonsogarciacaro:
Not sure why Travis is failing here. Any idea @forki?

forki:
Says yarn crashed

alfonsogarciacaro:
Argh, yarn is giving a lot of problems in Travis 😁 @forki Should we use npm 5 which also has lock files?

forki:
I'm not allowed to comment on package managers. My karma is about to enter negative numbers

mastoj:
I wonder if this was the same issue I mentioned for you at NDC @forki ? I did a run and it said `yarn` was missing, added a stupid line that printed some text and it all started working as expected. I have no clue what was broken or what fixed it.

alfonsogarciacaro:
Apparently the problem was adding the `phantomjs` dependency to package.json. I really have no idea how that line got there 😕 

After removing it, tests pass so I'm merging :+1:

PR #125, opened 2017-06-21
Convert into a template

MangelMaxime:
In this PR, the template should be ready to publish.

We probably need some automation to generate it or publish it. If someone is willing to do it would be appreciate. I will not have time to do it this week :)

Would be nice if you can review the [template info](https://github.com/fable-compiler/fable-suave-scaffold/pull/125/files#diff-fe7aba133e1fcd723001de9a7ae59093) and give the template a test locally before we publish it.

It's running on my computer but we are never too careful :)

Edit: cc @bentayloruk If you want to start working on updates you can start from this branch. And create another PR I believe.


MangelMaxime:
To fix the tests, should we run them on the projects located into: `Content/` ?

alfonsogarciacaro:
@enricosada Sorry, could please explain what do we need to do in the .proj file to pack the template so we don't need to move everything to the Content folder? Thanks!

MangelMaxime:
I moved everything into the content because otherwise we have to choose which project we want to restore/pack. 

As we will have the `Tests.sln` and `*.proj` file at the same level.

This would not be a problem if we used a build.fsx for example to pack/restore/publish but in this case we have "2" build.fsx at the same level :p

Also, I prefer to split the repo specific files from the template files. I find it easier to read than asking yourself, which files will be including into the templates ?

All a mater of personal choice :) Otherwise, if you want to not use the content folder, you simply need to adapt the `include` and `exclude` information [here](https://github.com/fable-compiler/fable-suave-scaffold/pull/125/files#diff-ffc892a392d5369f6dd0132f2a1c593aR19)

bentayloruk:
Thanks @MangelMaxime!  

How do I `build run` in this new world?  I tried `build run` in the `Content` folder, but the `.paket` folder has no `paket.exe` in it.  Is this an oversight, or is it not appropriate in a template?  Should I be running something to create an instance of the template?

I'd like to be able to `build run` and `build`, before I hack at anything, so pointers appreciated!

MangelMaxime:
@bentayloruk Should be fix.

enricosada:
> could please explain what do we need to do in the .proj file to pack the template so we don't need to move everything to the Content folder? Thanks

Why are you moving eveything in a `Content` folder?
I am not understanding that.

Is not easier to leave this repo as is, so `git clone` will work as normal? And add to content?
I dont understand what do you mean with `what do we need to do in the .proj file`. 

If the `template.proj` is in a subdirectory, maybe use relative path

```xml
 <Content Include="../" Exclude="$(ExcludeFromPackage)" >
          <PackagePath>Content\</PackagePath>
 </Content>
```

what is not working?

about fake script, you can have `build.fortemplate.fsx` and rename when you content include

```xml
 <Content Include="../build.fortemplate.fsx" >
          <PackagePath>Content\build.fsc</PackagePath>
 </Content>
```

so you have both in parallel, with clone is the development fake script, and packaged in template is the one needed (who can also be run as `fsi build.fortemplate.fsx` to try it

MangelMaxime:
Oh I didn't know we could do this manipulation. Thanks for the info @enricosada will contact you on slack if I am stuck :)

I will rework this PR so we don't have a `Content` folder.

forki:
closing for now. Please resubmit when done

MangelMaxime:
Thanks forki

PR #127, opened 2017-06-28
Fix #126 by using newUrl rather than modifyUrl.

### LINKED ISSUE #126: Login is not shown
I have a rather small problem, after logging out the second time, I need to click on home before the login screen is shown.

So the steps to reproduce:
click “Logout”, -> Welcome page is shown
click “Login” -> Login page shown, logging in
click “Logout” -> click Login welcome page is still shown 
I have to click “Home” and then “Login” again to display the Login page.

If I open a new page and open localhost:8080, it is also after the second “Logout”, that I have to press “home” before the login is displayed.

I can reproduce it with Chrome Version 59.0.3071.115  (64-Bit)
and IE Version 11.0.9600.18665.

It is not big issue for me, but I know some people who think,  if that does not work what is with the rest.

bentayloruk:


alfonsogarciacaro:
Thanks a lot for investigating and fixing this @bentayloruk!

Issue #128, opened 2017-07-01
Unit tests missing


aprooks:
Since it's  scaffold I'd expect it to have unit tests projects separate from UITests.

I could add backend test example via expecto but no idea how to deal with it on Fable side. 

What do you think?

alfonsogarciacaro:
We want to create a test library/layer for Fable with an API similar to Expecto but it's not done yet. However, it's not difficult to use any JS test library with F# code, see for example [Fable.PowerPack tests](https://github.com/fable-compiler/fable-powerpack/blob/b035313b7376922446d51e503ef2913ef41d2172/tests/PromiseTests.fs#L9-L32).

aprooks:
Thanks, I will start with [easy] backend-side UT for now.

aprooks:
I seem to have found a way of running tests and server via single watch: https://github.com/aspnet/DotNetTools/pull/275/files#diff-a055dfae715500c275a4f05108c9362c will try it later.

aprooks:
Server side is fine. I will need more time to work with Fable and sort out some good example.

I was thinking of making a starter test on model transitions like @forki  mentioned on his talks about React native. Something like 

```
Given Model
|> When message 
|> Then (fun (cmd,changedModel) ->
    assert.equals changedModel.Page  NewPage
)
```

PR #129, opened 2017-07-04
server tests starter with live reload

aprooks:
starter for #128 

what's working: live reload of test project and tests run.

I did not use Expecto Fake helpers, not sure if they would work with 'dotnet watch'

also I noticed at least once that 2 watching process deadlocked. Not sure what to do with it atm.

Signed-off-by: Alexander Prooks <aprooks@live.ru>

### DIFF
--- a/test/ServerTests/Tests.fs
+++ b/test/ServerTests/Tests.fs
@@ -3,7 +3,6 @@ module ServerTests.Tests
 open Expecto
 open ServerCode
 
-[<Tests>]
 let wishListTests =
   testList "Wishlist" [
     testCase "default contains F# mastering book" <| fun _ ->
--- a/test/ServerTests/Tests.fs
+++ b/test/ServerTests/Tests.fs
@@ -1,0 +1,15 @@
+module ServerTests.Tests
+
+open Expecto
+open ServerCode
+
+[<Tests>]
+let wishListTests =
+  testList "Wishlist" [
+    testCase "default contains F# mastering book" <| fun _ ->
+      let defaults =  WishList.defaultWishList "test"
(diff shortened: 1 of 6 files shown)


forki:
Can't believe we are already at #129 here. ;-) this was just supposed to be a small example. But seems people are interested. Really cool tbh

aprooks:
* added serverTests project building on default build
* added running tests via `dotnet watch`

things to note. 

I've made 2 separate tasks for .net framework executables and for running serverTests project via dotnet
also `RunServerTests` task is put earlier than 'ClientBuild' since I prefer fail fast approach during builds and tests are being completed much faster than fable compilation

aprooks:
> Can't believe we are already at #129 here. ;-) this was just supposed to be a small example. But seems people are interested. Really cool tbh

small examples are very important imho :) I prefer to clone/download full working example and then adjust/delete example code instead of setting up solution from the ground every time

aprooks:
it is working for me now ;)

alfonsogarciacaro:
Nice work @aprooks! This looks good to me to merge 😄 Do you have any comment, @forki?

test/ServerTests/Tests.fs
```diff
@@ -0,0 +1,15 @@
+module ServerTests.Tests
+
+open Expecto
+open ServerCode
+
+[<Tests>]
```
forki: Please use expecto tests without the attribute. I think we don't need that. Thx

aprooks: thanks, dunno where it came from :)

aprooks: @forki  removed

forki:
:kiss:

PR #131, opened 2017-07-05
fix comment

forki:


alfonsogarciacaro:
Actually, although Fable.Core 1.1.6 and dotnet-fable 1.1.6 have been also released in band, as we're quite stable now, I've made it so it's not necessary to release a new Fable.core version for every tiny change in the compiler. So it's shouldn't be necessary to keep both versions in sync anymore.

However, just in case, I'm still forcing to use a specific Fable.Core version so user needs to take care that both core and compiler are updated to the latest versions. This means we still have the problem to coordinate versions across paket.dependencies and .fsproj, but at least users shouldn't have to pin Fable.Core version in general.

forki:
@alfonsogarciacaro please come on gitter. I want to solve it

PR #132, opened 2017-07-05
Use CLI magic

forki:
depends on release of https://github.com/fsprojects/Paket/pull/2486


forki:
Note: this should only be merged when paket 5.5 is released. and we should not merge the paket config file

forki:
OK changed my mind - I will merge this now and it will use the paket 5.5 rc1. After release we go back on stable channel. But I really want people to test it

PR #134, opened 2017-07-07
single watch project for backend test/run

aprooks:
adjusted Tests project and Build script so that whole backend rebuild, tests execution and server start is run sequentially. 

Fixes deadlock due to parallel build when run via 2 watcher scripts of server and tests.

main issue here is that Logs colors are lost due to processes being wrapped by MsBuild

Following screenshot shows what happens when you change server code, then fix UT:

conemu64_2017-07-07_22-37-21

if colors are critical I will add new target to run server only.  Tests will have to be run separately and manually


alfonsogarciacaro:
I haven't used it yet but there's a way to run Expecto tests with Ionide and see a marker for passing/failing tests directly in the editor. Maybe we could recommend that? @Krzysztof-Cieslak

aprooks:
I could not make it work with project targeting netstandard.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

alfonsogarciacaro:
Ah, ok, thanks. Then we should merge this, I'll wait for @forki comments :)

PR #135, opened 2017-07-13
Add debugger config

Krzysztof-Cieslak:
596760f812e1b998067789

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Debug",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "Start",
+            "program": "${workspaceRoot}/src/Server/bin/Debug/netcoreapp1.1/Server.dll",
+            "args": ["${workspaceRoot}/src/Client"],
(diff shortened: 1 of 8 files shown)


Krzysztof-Cieslak:
I've added text in readme ;P

forki:
💋

.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+            "cwd": "${workspaceRoot}",
+            "stopAtEntry": false,
+            "console": "internalConsole"
+        },
+        {
+            "name": "Debug Server Test",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "Build ServerTests",
+            "program": "${workspaceRoot}/test/ServerTests/bin/Debug/netcoreapp1.1/ServerTests.dll",
```
forki: somehow we should get rid of this here!?

Krzysztof-Cieslak: Not really sure how, that's just how Code debugger configuration works

Krzysztof-Cieslak: How it should look like? Maybe we can add some hack around it

forki: yeah I have no idea. it's a bit weird ;-(

PR #136, opened 2017-07-15
[WIP] Update everything to sdk 2.0

forki:
@enricosada @alfonsogarciacaro @cartermp @haf @kevinransom @Krzysztof-Cieslak @dsyme 

Here we are trying to move something that is used in the real world to the new sdk model and netcore 2.
I assume we need to work together on this and iterate on it. There are many chaning parts here: ionide, VS, Dotnet sdk, paket, ....

If someone sees things that needs to be changed then please send a PR to this or just comment here.
Thanks


forki:
/home/travis/build/fable-compiler/fable-suave-scaffold/src/Server/Server.fsproj : warning NU1605: Detected package downgrade: Microsoft.NETCore.Platforms from 2.0.0-preview2-25405-01 to 1.1. Reference the package directly from the project to select a different version. 


any ideas what that means? or why?

alfonsogarciacaro:
I don't know what's the status of netcore 2 & F# at the moment. Is is worth already starting testing it with real projects? Should we wait at least until preview3?

alfonsogarciacaro:
@forki Not sure if it's exactly the same problem, but once I got something similar if I didn't add some explicit references to dotnet-fable [check this](https://github.com/fable-compiler/Fable/blob/0ae3ec1bb1d5dc9ac0ee1ea911a2d686f20ef4e7/paket.dependencies#L12-L15).

forki:
The message is against the netcore projects in server - not the fable client stuff. 

Regarding "is it worth it": if we don't report errors now then we will end up with something that does not work. Nobody is going to test it for us. We need to do it.

cartermp:
A few things to answer @alfonsogarciacaro's questions:

1. .NET Core 2.0 + F# is stable enough to scout out this change now and hash out the issues.

2. To test with the right bits, you'll need to do this against the [latest coherent build](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/release/2.0.0/latest.coherent.version).  This is the version that is intended to work with the runtime, ASP.NET, etc.   There's currently no direct link to this build, so you'll have to hunt it down 😢.

3. I'd like to have build -> run -> publish of this project be a part of Microsoft's manual scenario validation of the CLI, SDK, and .NET Core runtime.  I currently have a manual validation script that I have given to vendors who walk through the steps and report back if there were any concerns.  They do this for all of .NET Core 2.0.  This is the fastest way to identify problems when they arise.  When vendors report back a problem, I can create an issue here and we can then figure out where the bug lies.  This will continue as the runtime evolves.  For example, [2.1 is already being worked on](https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/latest.coherent.version).  Getting an understanding of any breaks earlier in advance should make things a lot easier on you.

forki:
@cartermp can you find the download page for that thing. It's a bit unfortunate that latest versions have guids in the urls. This makes it extremely hard for FAKE

cartermp:
I tried to find it a bit this afternoon and couldn't.  There's no download page for the latest coherent builds, unfortunately.  I'll ask on Monday when folks are in the office if I can't get it this weekend.

forki:
Thanks. And please let them know that earlier all links were systematic. That was a good time

alfonsogarciacaro:
@cartermp Unfortunately this is what I meant by mentioning that preview2 is not mature enough and it might be better to wait for an _official_ preview3. I'm afraid the community effort will be much less effective if we've to spend a lot of time to make sure everybody is hunting down the right version in order reproduce issues in our machines, in other operating systems... @forki and other members in Fable community have worked hard to make sure we could have reproducible builds in Fable projects, but all of this becomes nothing if we don't know which SDK version people are using 😕

cartermp:
Still waiting to hear back.  Let's keep this PR open until Preview 3 bits are available.  I think it's in a weird state right now - one small, but important, change to F# support is in the release branch and not Preview 2.  I'm not sure if it would affect FABLE, but I don't think this is worth testing on Preview 3 given that we do have some upstream changes coming in.

This effort that I'm talking about is also more long-term.  Core 2.0 is intended to be a more stable baseline that 1.0 ever was, so once the support for FABLE on top of it has been established, it should be relatively easy to include in our manual scenario validation.

cartermp:
Hello!  Can't believe I forgot about this until just now.  There was no preview 3, after all 😄.  Development on the 2.1 release is underway.  Is there interest in getting SDK 2.0 support?  Once that is in place, I can get building this repository in a manual test suite that will be run on a daily basis for .NET Core.

alfonsogarciacaro:
@cartermp I published a prerelease of Fable 1.2 with netcore2 support so I've updated this PR to use it (and other fixes for netcore2 compatibility). It seems to work so I'll merge the PR. Please feel free to add this repository to the test suite :)

cartermp:
Danke shurne!

cartermp:
Got this added to our manual validation for .NET Core!  Our vendors will now manually verify this simple process on a weekly cadence:

Clone --> run the build script and verify everything builds and runs --> edit an F# file and verify that the code recompiles and the app changes in the browser

PR #140, opened 2017-07-15
apply FrameworkPathOverride in the build script

baronfel:
otherwise we have to depend on every user setting it.


baronfel:
Man, it'd be great if travis came with a pre-installed dotnet sdk. That initial package cache download is a pain.

baronfel:
ok, I moved the variable into the build.fsx

forki:
:kiss:

PR #146, opened 2017-08-18
Ensure scaffold will still build with Suave 2.2.x

### LINKED ISSUE #145: Minor change needed for scaffold to build against Suave 2.2.x
### Description

The recent release of Suave 2.2.x includes a [breaking change](https://github.com/SuaveIO/suave/commit/2375fdd5da6289a09ec35359d12bc38420d31c54) that, among other things, hides the `YoLo.fs` file from user code (so its functions are only exposed internally to Suave, and not to end-user code that consumes the Suave library).

This causes one line of the scaffold to break: in [`Auth.fs`](https://github.com/fable-compiler/fable-suave-scaffold/blob/7defd4d7380b7fcff60a4afd8c6d628feb90ce72/src/Server/Auth.fs#L9), there's a function that uses `UTF8.bytes`:

```fsharp
let UNAUTHORIZED s = unauthorized (UTF8.bytes s)
```

`UTF8.bytes` comes from `YoLo.fs`, so it's not accessible once you move to Suave 2.2.0 or later. The fix is to replace that line with:

```fsharp
let UNAUTHORIZED (s : string) = unauthorized (System.Text.Encoding.UTF8.GetBytes s)
```

Alternately, since the `unauthorized` and `UNAUTHORIZED` functions are already defined in `Suave.RequestErrors`, they could simply be deleted from `Auth.fs`, making the solution even simpler.

rmunn:
The UTF8.bytes function is hidden in Suave 2.2.0 and above, so we can't use it in the scaffold. Since the only functions that use it are exact duplicates of functions already in the Suave API, just remove them.

This will fix #145.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -148,7 +148,7 @@ Target "RenameDrivers" (fun _ ->
     try
         if isMacOS && not <| File.Exists "test/UITests/bin/Debug/net461/chromedriver" then
             Fake.FileHelper.Rename "test/UITests/bin/Debug/net461/chromedriver" "test/UITests/bin/Debug/net461/chromedriver_macOS"
-        elif isLinux then
+        elif isLinux && not <| File.Exists "test/UITests/bin/Debug/net461/chromedriver" then
             Fake.FileHelper.Rename "test/UITests/bin/Debug/net461/chromedriver" "test/UITests/bin/Debug/net461/chromedriver_linux64"
     with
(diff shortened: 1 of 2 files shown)


rmunn:
Also pushed a fix for #147 while I was at it, so that both bugs can be fixed by merging the same PR.

alfonsogarciacaro:
Thanks for the PR and sorry for the late response!

PR #162, opened 2017-09-18
Update deps

forki:


forki:
@alfonsogarciacaro urgs - fable update broke it!

PR #164, opened 2017-09-21
Update readme.

isaacabraham:
Also update book url :-)


src/Server/WishList.fs
```diff
@@ -24,9 +24,9 @@ let defaultWishList userName : WishList =
             [{ Title = "Mastering F#"
                Authors = "Alfonso Garcia-Caro Nunez"
                Link = "https://www.amazon.com/Mastering-F-Alfonso-Garcia-Caro-Nunez-ebook/dp/B01M112LR9" }
-             { Title = "Learn F#"
+             { Title = "Get Programming with F#"
```
forki: sorry we need an unit test for that!

https://github.com/fable-compiler/fable-suave-scaffold/blob/master/test/ServerTests/Tests.fs#L12

:P

PR #180, opened 2017-09-28
Update to react 16

forki:


alfonsogarciacaro:
Any noticeable changes with React 16? I guess nothing broke in the bindings, right?

forki:
the website says if 15.x works then 16 should work as well

PR #181, opened 2017-09-28
Update to react 16

forki:
Urgs: TIL this ugliness. How can we make it use the real version that we have in package json

### DIFF
--- a/src/Client/index.html
+++ b/src/Client/index.html
@@ -7,10 +7,6 @@
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap-theme.min.css">	
 		<link rel="stylesheet" href="css/site.css">
 		<link rel="shortcut icon" type="image/png" href="/images/safe_favicon.png"/>
-
-		<!-- React -->
-		<script src="https://cdnjs.cloudflare.com/ajax/libs/react/16.0.0/react.min.js"></script>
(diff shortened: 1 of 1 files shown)


src/Client/index.html
```diff
@@ -9,8 +9,8 @@
 		<link rel="shortcut icon" type="image/png" href="/images/safe_favicon.png"/>
 
 		<!-- React -->
-		<script src="https://cdnjs.cloudflare.com/ajax/libs/react/15.4.0/react.min.js"></script>
-		<script src="https://cdnjs.cloudflare.com/ajax/libs/react/15.4.0/react-dom.min.js"></script>	
+		<script src="https://cdnjs.cloudflare.com/ajax/libs/react/16.0.0/react.min.js"></script>
```
forki: @alfonsogarciacaro @MangelMaxime what's the fablish/webpackish way to deal with that?

MangelMaxime: @forki You should not need to use react from cdn.

Your webpack.config.js doesn't list `react` as an external dependencies so it's should already be inside your `bundle.js`.

forki: thanks. is that the correct way?

MangelMaxime: I would say yes.


If you want to benefit from cdn cache on your application you can probably explore webpack plugins like [dynamic-cdn-webpack-plugin](https://github.com/mastilver/dynamic-cdn-webpack-plugin).

Which will use CDN for your deps on production env. And will choose the same version as the one used in your package.json

But I didn't tested it.

PR #182, opened 2017-09-28
Add Hot Module Reloading as direct file

forki:
@MangelMaxime no pressure - just trying if it works here


MangelMaxime:
@forki Ahah

No problem. Checking HMR on several projects before making it open, can only be something good.

forki:
ok. what's the easiest way to see if it really works and it's not just a reload?

MangelMaxime:
If you have a form go to it. Set some values into the form's input and update one of the form's label.

If you see the label updated and your form keep the values inside the input. HMR is working.

You can also, check the console of your browser. If the page is reloaded, then the console is almost empty.

forki:
ok then it's not working

MangelMaxime:
@forki You need to update your webpack.config.js with this changes:

https://github.com/fable-elmish/templates/blob/e5ea09321ef67d710801273e0beda899736116ae/src/react-demo/Content/webpack.config.js#L38-L39

https://github.com/fable-elmish/templates/blob/e5ea09321ef67d710801273e0beda899736116ae/src/react-demo/Content/webpack.config.js#L71-L74

forki:
hmr

MangelMaxime:
So it's working ? You know that I am stressing with all your tests 😄

forki:
it's working like a charme

forki:
This is it on mobile. App continues to run over recompile and keeps all state.

hmrmobile

src/Client/webpack.config.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
   devtool: "source-map",
   entry: resolve('./Client.fsproj'),
   output: {
-    path: resolve('./public'),
+    path: resolve('/public'),
```
alfonsogarciacaro: I think this is going to resolve `public` in the root path of your drive (e.g. `C:\public`), you probably want to keep the period or remove the slash.

src/Client/App.fs
```diff
@@ -181,9 +182,10 @@ open Elmish.Debug
 // App
 Program.mkProgram init update view
 |> Program.toNavigable (parseHash pageParser) urlUpdate
-|> Program.withConsoleTrace
-|> Program.withReact "elmish-app"
 #if DEBUG
+|> Program.withConsoleTrace
 |> Program.withDebugger
+|> Program.withHMR
```
alfonsogarciacaro: Didn't we agree to name this `Program.withAllTheHotThingsAvailable`? :wink:

src/Client/webpack.config.js
MangelMaxime: I will keep the `.` then :)

src/Client/App.fs
forki: look at the commit message here. Might bring inspiration

alfonsogarciacaro: To hot to be true 😆

forki:
anyone seen this before:

MangelMaxime:
@forki Do you see this since you add Elmish.HMR ?

forki:
Maybe unrelated and just because of paket update?

Am 28.09.2017 17:40 schrieb "Maxime Mangel" <notifications@github.com>:

MangelMaxime:
Perhaps, because this updated dotnet-fable and F# compiler version too.

Perhaps @alfonsogarciacaro know something about that.

alfonsogarciacaro:
Ups, [I marked `applySpread` as Obsolete](https://github.com/fable-compiler/Fable/blob/71652e074d870edb2ad62f22366d8c2905e24b6e/src/dotnet/Fable.Core/Fable.Core.JsInterop.fs#L53-L54) but the React helpers are still using it, I have to publish a new version, sorry! 😅

forki:
Cool. So dogfooding works....

Am 28.09.2017 17:51 schrieb "Alfonso Garcia-Caro" <notifications@github.com

alfonsogarciacaro:
Here #184

forki:
I think it also broke RN package:

image

forki:
image

vasily-kirichenko:
If I wrote this:
```fsharp
...
|> Program.withHMR
|> Program.withReact "elmish-app"
```
It works nicely.
If I wrote this:
```
...
|> Program.withReact "elmish-app"
|> Program.withHMR
```
It does not update the current page automatically.

I think it's a not very good at all. It's very implicit and hard to spot out.

forki:
yes. known issue. @MangelMaxime did you document that? maybe we can throw if we detect something?

MangelMaxime:
@vasily-kirichenko We have no choice here in fact. This is documented.

And in fact, if you use:

```
|> Program.withReact "elmish-app"
|> Program.withHMR
```

HMR is working, this is just Elmish.React not re-rendering the view because the model didn't change. If you go to another page and go back you will see the changes. This is because with are using `lazyView` to make rendering more performant.

MangelMaxime:
For react, you would need to manually specify what to re-render like they do [here](https://github.com/gaearon/react-hot-boilerplate/blob/next/src/index.js#L16)

I don't find it really user friendly neither. I don't think I can check the declaration order.

forki:
@MangelMaxime @alfonsogarciacaro ok I fixed Fable.ReactNative

et1975:
> Yeah, I see. I just think that it's not a good "fluent" API if shuffling functions in a pipe changes behavior. 

Yep, It is out of control. Grew out of just a couple of functions... now there's certain order requirements for everything:  navigation, debugger, react and now HMR :/ I'm open to suggestions.

isaacabraham:
Wrap the result in a DU / custom record that enforces the order through the type system?

forki:
these are different project/nuget packages. not easy to do

isaacabraham:
What about some validation on one that checked the state of the context and throws if it's not set up correctly? Better than nothing and better than runtime unexpected behaviour.

alfonsogarciacaro:
As this will run before anything, I guess it's not that necessary to use the type system because the user can see an error in the runtime very soon. Maybe add a set of tags to the program, each helpers adds its own tag so the next one can check if any incompatible function has been run beforehand.

et1975:
I don't know, I still feel it's akin to switching the order of `concat` args and expecting the same result. 
I think I'd like to solve it through types or not at all.

Issue #189, opened 2017-10-02
Bundle shim.min.js and fetch.js with webpack


kerams:
Is it necessary to keep the 2 files in the project rather than adding them to `packages.json`?

See `index.html`
```html
<script src="js/shim.min.js"></script>
<script src="js/fetch.js"></script>
```

forki:
@MangelMaxime @alfonsogarciacaro how can we bundle it? and do we want to?

kerams:
Fetch https://www.npmjs.com/package/whatwg-fetch
Shim https://www.npmjs.com/package/core-js

I'm not sure whether Webpack would boil the whole core-js file down to the minified version at https://cdnjs.com/libraries/core-js/2.4.1

kerams:
Incidentally, `yarn.lock` already contains both of these, so we could just remove them?

MangelMaxime:
@forki For fetch I think doing: `JsInterop.importAll "whatwg-fetch"` works.

We have it in [this sample](https://github.com/fable-elmish/sample-react-navigation/blob/master/src/app.fs#L16)

forki:
ok what about the other one?

MangelMaxime:
For `shim.min.js`, I am not sure but I think we can just remove it completely.

We are using babel [here](https://github.com/SAFE-Stack/SAFE-BookStore/blob/master/src/Client/webpack.config.js#L9-L12) which should already include everything needed to support ES2015.

Also, we should now use [babel-env](https://babeljs.io/docs/plugins/preset-env/) from now as presets has been deprecated.

forki:
can you send the PR for babel-env?
I will remove the shim

forki:
@MangelMaxime https://github.com/SAFE-Stack/SAFE-BookStore/pull/192 looks like removing it breaks the build

MangelMaxime:
@forki Sure, I will do

MangelMaxime:
@forki Hum, I will take a look about the shim stuff. Because, I use the same stack as this project without shim and the UITests were working last time I run them.

MangelMaxime:
@forki I can't run the UITest suites on my computer because it's complain about missing references for `.NETFramework,Version=v4.6.1` but when I try to install .Net4.6.1 it's say that I have it or newer... All fun...

[...]

PR #191, opened 2017-10-02
Import fetch - references #189

forki:
/cc @MangelMaxime @kerams


src/Client/App.fs
```diff
@@ -14,6 +14,8 @@ open Client.Messages
 open Elmish.Browser.UrlParser
 open Elmish.HMR
 
+JsInterop.importAll "whatwg-fetch"
```
alfonsogarciacaro: There's also `JsInterop.importSideEffects "whatwg-fetch"` for importing something without assigning it into a variable. In this case it has the same effect, just maybe the semantics is a bit clearer.

PR #195, opened 2017-10-02
Use babel-polyfill

MangelMaxime:


forki:
Looking good

MangelMaxime:
Yes, just waiting for travis to start. :)

forki:
Argh I was too fast. Damn it

Am 02.10.2017 17:21 schrieb "Maxime Mangel" <notifications@github.com>:

MangelMaxime:
@forki Travis failed but it's doesn't seems to be caused by this PR. I would says connection problems perhaps ?

alfonsogarciacaro:
I wonder why `babel-polyfill` is need and the `transform-runtime` plugin is not enough 🤔 Maybe we could use [polyfill.io](https://polyfill.io/v2/docs/) as @MangelMaxime did for some projects. It also seems to be possible to add polyfills in Webpack entry which may be a bit cleaner:

```js
module.exports = {
  entry: ["whatwg-fetch", "babel-polyfill", "./src/MyProj.fsproj"]
};
```

forki:
Happy to get PRs ;-)

Am 02.10.2017 21:59 schrieb "Alfonso Garcia-Caro" <notifications@github.com

MangelMaxime:
@alfonsogarciacaro As `transform-runtime` is not enough is it it safe to remove it if we use `babel-polyfill` ?

According to the doc, I guess they are both doing the same work here: https://babeljs.io/docs/usage/polyfill/#details

@forki I am testing using webpack to manage the polyfill inside another repo. If my experimentation works, I will do a PR for BookStore.

alfonsogarciacaro:
@MangelMaxime Yeah, I think so. IIRC the difference is `transform-runtime` _injects_ the polyfills in the code while `babel-polyfill` polyfills everything at once. In theory the former should be a bit more performant because it only adds the polyfills you actually use to the bundle.

BTW, I checked in the documentation that `babel-polyfill` is a combination of regenerator runtime and core-js. Actually, we should only need `core-js` as Fable generates its own sequences (Iterable in JS).

PR #200, opened 2017-10-04
Support for Azure Tables

isaacabraham:
Adds pluggable support for Azure Tables (defaults to file system) through a command line argument e.g.

 `dotnet run -- AzureConnection=UseDevelopmentStorage=true`

If you don't want to use a live Azure account, you can install the latest Azure Storage Emulator [here](https://azure.microsoft.com/en-us/downloads/) - scroll to the bottom and download the Storage Emulator (windows only). Also consider downloading Azure Storage Explorer at the bottom of the page (x-plat).


forki:
ok so how do we configure to use azure? Because: `dotnet run -- AzureConnection=UseDevelopmentStorage=true` won't really work here.

In this sampe we don't run the dotnet cli manually and I'd like to use the azure thing for the demo that runs on azure in the container at http://fable-suave.azurewebsites.net/#home

can we do that?

isaacabraham:
How does the app get launched within the container? As long as we can pass some cmdline arguments through, it should work.

Regarding the table storage itself, we have two options:

1. Host the development emulator within the container (but this then means windows-only).
2. We create a storage account alongside the container service and configure the container service to use the storage account. I assume there's some way to securely configure the container with "configuration" details like this (connection strings etc.)?

forki:
oh yes I can launch it with that param on azure. but locally it's a bit different workflow.

> We create a storage account alongside the container service and configure the container service to use the storage account

yes that's what I want

forki:
how is that storage account called exactly? so that I find it in the wodnerful azure portal

isaacabraham:
An [Azure Storage](https://azure.microsoft.com/en-us/services/storage/) account. Once we have this added and we know what the infrastructure looks like e.g. Azure Storage account + Azure App Service w/ Docker, we can look at creating an ARM template that will provision everything in one go.

forki:
image

PR #204, opened 2017-10-08
Auto reset Azure DB every 1 hour.

isaacabraham:


.vscode/launch.json
```diff
@@ -5,14 +5,25 @@
             "name": "Debug",
             "type": "coreclr",
             "request": "launch",
-            "preLaunchTask": "Start",
+            "preLaunchTask": "Build Server",
```
forki: Cc @Krzysztof-Cieslak

isaacabraham: Start didn't work for me when called through Code's debug / launch. But happy for that to be rolled back if needed.

forki:
Can you please set the last db reset time as one mutable field or into the db? So that we can show that in UI

isaacabraham:
@forki what should it give for the file system version? None?

forki:
App startup?

Am 08.10.2017 15:54 schrieb "Isaac Abraham" <notifications@github.com>:

isaacabraham:
@forki done - can you test it out locally (you can change local time to e.g. every 10 seconds in `Program.fs`.

isaacabraham:
BTW: When type providers are properly supported and we port across the Azure TP, this code will be much nicer :-)

forki:
Will do tomorrow. Thanks

PR #208, opened 2017-10-09
Removed usage of lazyView2

purkhusid:
This change is intended for better first time experience with HMR. Currently lazy views are not invalidated and reloaded with HMR unless you make changes to the state of the component or change the model of the component. This makes the first time experience with HMR really confusing.


MangelMaxime:
I never used lazyView in my application yet.

In general, you use them to improve the performance of the rendering because you only make the diff if the requested portion of the model changed.

PR #213, opened 2017-10-11
Share API endpoint URLs between server and client

kerams:


forki:
that is soooo coool. Isomorphic F# FTW

forki:
giphy

enricosada:
Isomorphic is so old schoold. hot buzz now is `Universal`.
Let's say it is an Universal Web Project => UWP.

forki:
I like how "UWP" sounds. @Krzysztof-Cieslak will love it

Krzysztof-Cieslak:
giphy-downsized

tforkmann:
nice feature!!

panesofglass:
This isn't quite Universal, is it? Isn't Universal when the same front-end urls are used, not just the same api urls?

forki:
@kerams 

I have 

    pathScan "/api/books/%s/" (fun key -> Books.getBook key)

how would we do the Url sharing?

forki:
@panesofglass you are talking about something else

kerams:
@forki Interesting question. I tried storing the URL as `PrintfFormat` and `TextWriterFormat`, but I always ended up with some combination of losing type safety/inference, verbosity and clunkiness, or having 2 copies of the URL :).

et1975:
I'm clearly late to the party, but imho this is an anti-pattern.  I think we need to strive towards [higher API maturity levels](https://martinfowler.com/articles/richardsonMaturityModel.html) instead.

forki:
Ok I'm all ears. So what would be a good way out?

Am 26.10.2017 19:35 schrieb "Eugene Tolmachev" <notifications@github.com>:

isaacabraham:
Maybe some proposal PR would be great. Even something that doesn't compile, just something that shows the idea you're thinking of.

et1975:
I haven't done anything with custom mime types, but it's quite possible to avoid sharing the api links (aka lvl3) - all you have to do is add the server-side links every time you want to communicate a possible transition.  For example, when your `/books` api returns a book, it also returns available "controls" - links to do something with that book:
``` 
{
 books: [{"id":123, ..., links: [{"del","DELETE:/book/123"}] }],
 links: [{"add","POST:/books"}]
} 
```
Your server ends up fully in control of the REST api (but is also able to guide the client towards actions available at the moment) and all the client needs is root of the api and a way to "interpret" the links. The format of the links is entirely up to the implementation, for example I know which action (add,del,update) map to witch VERB so I don't add those (but it's in the example for clarity).

When you do something that adds a new resource, you also return a link, for example in the `POST:/book` implementation I'd return `links:[{"get","GET:/book/123"}]` and boom! You client knows where to find it if it wants to load it.

forki:
Yes I do understand it in principle. But from a implementation standpoint
I'm always a bit wondering.

Am 26.10.2017 20:12 schrieb "Eugene Tolmachev" <notifications@github.com>:

I haven't done anything with custom mime types, but it's quite possible to
avoid sharing the api links (aka lvl3) - all you have to do is add the
server-side links every time you want to communicate a possible transition.
For example, when your /books api returns a book, it also returns available
"controls" - links to do something with that book:

{
 books: [{"id":123, ..., links: [{"del","DELETE:/book/123"}] }],
 links: [{"add","POST:/books"}]
}

Your server ends up fully in control of the REST api (but is also able to
guide the client towards actions available at the moment) and all the
client needs is root of the api and a way to "interpret" the links. The
format of the links is entirely up to the implementation, for example I
know which action (add,del,update) map to witch VERB so I don't add those
(but it's in the example for clarity).

When you do something that adds a new resource, you also return a link, for
example in the POST:/book implementation I'd return
links:[{"get","GET:/book/123"}] and boom! You client knows where to find it
if it wants to load it.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/SAFE-Stack/SAFE-BookStore/pull/213#issuecomment-339752321>,
or mute the thread

PR #240, opened 2017-11-11
Move shared view menu.fs to root views folder.

bentayloruk:
`Menu.fs` is a shared view, rather than a page, so being in the `pages` folder is slightly confusing.  I suggest we have a root `views` folder for shared views like this.  I've developed an app from this template and that's what I found worked well.  It's also the approach used in the [Elm SPA example](https://github.com/rtfeldman/elm-spa-example).

If this is acceptable, I will amend the README where it refers to the current `Menu.fs` location.


forki:
Sounds good

Am 11.11.2017 14:08 schrieb "Ben Taylor" <notifications@github.com>:

bentayloruk:
I amended the commit in this PR with the README change, so it's good to go if wanted.

forki:
thanks

PR #248, opened 2017-11-30
Model variation

mexx:
I've played a little with the code to decouple the components as much as possible.

The user is now managed at the application level, as it was somehow awkward to access it from the menu component.
Menu is now simply a view without own state, as there is nothing to keep track of.
The login page state and messages are now only processed by the component itself, the embedding component get a notification about a successful login and can define how to react according to this fact.

Further work, not necessary in this PR, would be to separate the new book form as own component, similar to the login page.

Would love to hear your thoughts about the proposed changes.

### DIFF
--- a/src/Client/App.fs
+++ b/src/Client/App.fs
@@ -93,12 +93,7 @@ let update msg model =
 
     | LoggedIn newUser, _ ->
         let nextPage = Page.WishList
-        let m,cmd = urlUpdate (Some nextPage) { model with User = Some newUser }
-        match m.User with
-        | Some _ ->
-            m, Cmd.batch [cmd; Navigation.newUrl (toHash nextPage) ]
-        | None ->
-            m, Cmd.ofMsg (Logout ())
+        { model with User = Some newUser }, Navigation.newUrl (toHash nextPage)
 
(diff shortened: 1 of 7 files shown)


src/Client/App.fs
```diff
@@ -22,17 +23,15 @@ type PageModel =
...
-    | LoggedIn
+    | LoggedIn of UserData
     | LoggedOut
     | StorageFailure of exn
-    | OpenLogIn
-    | MenuMsg of Menu.Msg
     | LoginMsg of Login.Msg
     | WishListMsg of WishList.Msg
-    | Logout
+    | Logout of unit
```
kerams: What is the point of that `unit`?

kerams: Is it because you wanted to avoid `Menu.view (fun () -> dispatch Logout) model.User` down below?

mexx: Exactly, it's to be able to just `Menu.view (Logout >> dispatch) model.User`.

Same thing could be applied to `ClickLogIn` on login page. The usage would be `OnClick (ignore >> ClickLogIn >> dispatch)`.

Issue #235, opened 2017-11-04
Error while deploying on ACS


kunjee17:
### Description

I was following steps provided on ReadMe for Azure.

### Repro steps

Follow the steps on Read me for Azure other than Storage part.

### Expected behavior

It should run the sample application.

### Actual behavior

Getting following error.
```
502 - Web server received an invalid response while acting as a gateway or proxy server.
There is a problem with the page you are looking for, and it cannot be displayed. When the Web server (while acting as a gateway or proxy) contacted the upstream content server, it received an invalid response from the content server.
```

### Known workarounds

None. Try to redeploy docker image to my account but doesn't work

kunjee17:
I need to add `Server.dll` in start up file. I found out from image. (Should be added to readme?) 

Then getting new error `Service Unavailable` . I tried running docker locally but server is running on 8085 instead of 8080. Am I missing something?

forki:
Yes please add everything to the readme that was not clear.

Am 04.11.2017 05:09 schrieb "Kunjan" <notifications@github.com>:

I need to add Server.dll in start up file. I found out from image. (Should
be added to readme?)

Then getting new error Service Unavailable . I tried running docker locally
but server is running on 8085 instead of 8080. Am I missing something?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/SAFE-Stack/SAFE-BookStore/issues/235#issuecomment-341870306>,
or mute the thread

kunjee17:
@forki will update read me. What about port number? I guess because of suave running 8085, I was unable to access site even after giving Server.dll to start. What I should be doing there?

jeremyabbott:
I'm having the same problem... I found this: https://docs.microsoft.com/en-us/azure/app-service/containers/tutorial-custom-docker-image. I followed the instructions to set up the port, but it still doesn't work. I'm not sure what else needs to be done. You would also then need to map 443 to another port and expose another port in the docker file if you turned SSL on, but it's not clear where you would do that.

jeremyabbott:
tl;dr version:
`az webapp config appsettings set --resource-group <resource group name> --name <app service name> --settings WEBSITES_PORT=8085` needs to be run in the Azure CLI (easiest way is via Azure Cloud Shell from Azure Portal).

kunjee17:
[...]

PR #251, opened 2017-12-10
Adds ACS port mapping instructions to address issue #235

jeremyabbott:
This updates the readme with additional instructions for mapping the Azure App Service's port 80 to the containers port 8085. While testing this I noticed that it didn't appear to be necessary to specify a startup file.

If you want, I can also add the instructions for setting up the app using the Azure CLI:

Create the resource group
`az group create --name <resource group name> --location "West US"`

Create the app service plan
`az appservice plan create --name <app service plan name> --resource-group <resource group name> --sku S1 --is-linux`

Create the app service itself
`az webapp create --resource-group <resource group name> --plan <app service plan name> --name <app service name> --deployment-container-image-name <container image name>`

Map the app service's port 80 to the containers exposed port 8085
`az webapp config appsettings set --resource-group <resource group name> --name <app service name> --settings WEBSITES_PORT=8085`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # SAFE - A web stack designed for developer happiness
 
-The following document describes the [SAFE-Stack](https://safe-stack.github.io/) sample project. 
-SAFE is a technology stack that brings together several technologies into a single, coherent stack for typesafe, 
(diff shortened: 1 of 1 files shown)


forki:
Hugs

PR #255, opened 2017-12-15
First draft of Giraffe version

forki:


src/Server/Storage/AzureTable.fs
```diff
@@ -72,29 +74,32 @@ let saveWishListToDB connectionString wishList = async {
         operation
     
     let! booksTable = getBooksTable connectionString
-    return! booksTable.ExecuteBatchAsync batch |> Async.AwaitTask |> Async.Ignore }
+    let! _ = booksTable.ExecuteBatchAsync batch
```
gerardtoconnor: Were you not able to return the unit!? was type inference broken or compile/runtime error?

forki:
Is there a Task.Ignore?

Am 15.12.2017 22:58 schrieb "Gerard" <notifications@github.com>:

gerardtoconnor:
Not currently, do need one? How is it different to do!? Is it fact it can run (&ignore result) of task run outside of task CE?

alfonsogarciacaro:
Ups, actually we already had a working Giraffe version and were planning to send a PR 😅 You've been faster! It actually looks very close to this, I'll leave some comments.

BTW, do you want to PR to this repo or provide an alternative template/scaffold?

src/Server/Auth.fs
```diff
@@ -1,15 +1,13 @@
...
+open Microsoft.AspNetCore.Http
 
 /// Login web part that authenticates a user and returns a token in the HTTP body.
-let login (ctx: HttpContext) = async {
-    let login = 
-        ctx.request.rawForm 
-        |> System.Text.Encoding.UTF8.GetString
-        |> FableJson.ofJson<Domain.Login>
+let login next (ctx: HttpContext) = task {
+    let! login = FableJson.getJsonFromCtx<Domain.Login> ctx
```
alfonsogarciacaro: We're using the following helper (note it's asynchronous):

```fsharp
  let rawForm (req: HttpRequest) =
      use reader = new StreamReader(req.Body, System.Text.Encoding.UTF8)
      reader.ReadToEndAsync()
```
And then:

```fsharp
    let! reqData = Http.rawForm ctx.Request
    let login = FableJson.ofJson<Domain.Login> reqData
```

src/Server/FableJson.fs
```diff
@@ -14,3 +15,15 @@ let toJson value =
 
 let ofJson<'a> (json:string) : 'a =
     JsonConvert.DeserializeObject<'a>(json, [|jsonConverter|])
+
+let serialize x =
+    setStatusCode 200
+      >=> setHttpHeader "Content-Type" "application/json"
+      >=> setBodyAsString (toJson x)
```
alfonsogarciacaro: I actually like the Giraffe idea to make it a habit of using negotiation for the responses instead of always using `application/json`. But we can't use Giraffe's default `negotiate` because it doesn't use Fable JSON serialization so we have a custom one:

```fsharp
  let fableJson (dataObj : obj) : HttpHandler =
    setHttpHeader "Content-Type" "application/json"
    >=> setBodyAsString (FableJson.toJson dataObj)

  let negotiate (responseObj : obj) : HttpHandler =
    negotiateWith
        // Default negotiation rules
        (dict [
            "*/*"             , fableJson
            "application/json", fableJson
            "application/xml" , xml
            "text/xml"        , xml
            "text/plain"      , string >> text
        ])
        // Default unacceptable HttpHandler
        (fun (next : HttpFunc) (ctx : HttpContext) ->
            (setStatusCode 406
            >=> ((ctx.Request.Headers.["Accept"]).ToString()
                |> sprintf "%s is unacceptable by the server."
                |> text)) next ctx)
        // response object
        responseObj
```

Then, for example:

```fsharp
return! Successful.ok (Http.negotiate wishList) next ctx
```

src/Server/WishList.fs
```diff
@@ -1,49 +1,50 @@
...
+                    return! FableJson.serialize wishList next ctx
                 else
-                    return! BAD_REQUEST "WishList is not valid" ctx
+                    return! BAD_REQUEST "WishList is not valid" next ctx
         with exn ->
-            logger.error (eventX "Database not available" >> addExn exn)
-            return! SERVICE_UNAVAILABLE "Database not available" ctx
+            let msg = "Database not available"
+            let logger = ctx.GetLogger "wishlist"
+            logger.LogError (EventId(), exn, msg)
```
alfonsogarciacaro: Ah, so this is how you use the logger. I haven't activated it yet 😅

src/Server/Program.fs
```diff
@@ -42,7 +65,16 @@ let main args =
             |> Option.defaultValue Database.DatabaseType.FileSystem
 
         let port = getPortsOrDefault 8085us
-        WebServer.start database clientPath port
+        
+        WebHost
+            .CreateDefaultBuilder()
```
alfonsogarciacaro: We have:

```
    WebHostBuilder()
        .UseKestrel()
```

Is this the same? Asking out curiosity.

src/Server/Storage/AzureTable.fs
forki: yes Task.Ignore is missing

src/Server/Auth.fs
forki: it's same thing, right? mine just has it in one function

src/Server/FableJson.fs
forki: we should patch that into giraffe - or create a giraffe.fable or something

src/Server/Program.fs
forki: no idea.

src/Server/WishList.fs
forki: yes it's latest fix in girafe

src/Server/Auth.fs
```diff
@@ -18,20 +16,26 @@ let login (ctx: HttpContext) = async {
         let user : ServerTypes.UserRights = { UserName = login.UserName }
         let token = JsonWebToken.encode user
 
-        return! Successful.OK token ctx
+        return! Successful.OK token next ctx
```
forki: Ok one issue is that this is not working. It's not sending out the text but some weird stream with couple of the and f in it. @gerardtoconnor @alfonsogcnunez any ideas? Obvious fix is to put it in a record and use fablejson.serialize. But it's an issue that many will have coming from suave.

forki: It's also the issue why the pr is still red

src/Server/Program.fs
```diff
@@ -14,6 +20,23 @@ let getPortsOrDefault defaultVal =
     | null -> defaultVal
     | value -> value |> uint16
```
vasily-kirichenko: while you at it,
```fsharp
 let GetEnvVar var = 
     match Environment.GetEnvironmentVariable(var) with
     | null -> None
     | value -> Some value
```
=>
```fsharp
 let GetEnvVar var = Environment.GetEnvironmentVariable var |> Option.ofObj
```

src/Server/Program.fs
vasily-kirichenko: `CreateDefaultBuilder` is a helper that calls `UseKestrel`, `UseIISIntegration` and some other stuff.

vasily-kirichenko: https://github.com/aspnet/MetaPackages/blob/dev/src/Microsoft.AspNetCore/WebHost.cs#L148

vasily-kirichenko:
Do I understand right that .net core configuration capabilities are not used at all? It would be awesome to add an example of accessing `appsettings.json`. I uses a helper in my project:
```fsharp
GET >=> route "/api" >=> withConfig<MainConfig> >=> (fun (cfg: MainConfig) next ctx -> ...)
```
where `withConfig` calls `ctx.GetService<IConfiguration>().Get<'a>` (if I remember correctly).

It's cool because
* it's a part of .net core, so it works and will work
* it watches the config file and reread it
* it somehow bypass IIS caches so changes in the file immediately visible in the code (it's not the same if you read the file manually by `File.ReadAllText`; IIS caches it for about 2 minutes, which does not make life easier)

It's not cool because
* it's not typesafe

@colinbull wrote on Twitter he uses FSharp.Data.JsonProvider for typesafe `appsettings.json` access in a Giraffe project. It would by _very_ cool to add such an example to this repo.

forki:
What kind of config do we need? Example?

Am 16.12.2017 10:44 schrieb "Vasily Kirichenko" <notifications@github.com>:

vasily-kirichenko:
I dunno. Maybe Azure tokens or something like that? If this project does not need any config, then ok, forget about it.

forki:
No I'm really interested. We just need to find something "interesting"

Am 16.12.2017 11:00 schrieb "Vasily Kirichenko" <notifications@github.com>:

vasily-kirichenko:
I think it does not matter what we configure, this project should contain configuration access as an example.

forki:
I agree. Can you please come up with a pr? Would make it easier

Am 16.12.2017 11:06 schrieb "Vasily Kirichenko" <notifications@github.com>:

src/Server/Program.fs
alfonsogarciacaro: Thanks, @vasily-kirichenko! Do you know if `UseIISIntegration` has any effect when you run the server in a no-Windows platform?

src/Server/Program.fs
```diff
@@ -42,7 +65,16 @@ let main args =
             |> Option.defaultValue Database.DatabaseType.FileSystem
 
         let port = getPortsOrDefault 8085us
-        WebServer.start database clientPath port
+        
+        WebHost
+            .CreateDefaultBuilder()
+            .UseWebRoot(clientPath)
+            .UseContentRoot(clientPath)
```
alfonsogarciacaro: I forgot about this. I think ContentRoot should be different from WebRoot (= clientPath). In our case we're using the repo root. Although I haven't seen so far any use of ContentRoot yet, as we save the root in another value for file system access, etc. I guess it's used by ASP.NET services.

src/Server/Auth.fs
alfonsogarciacaro: Yeah, I just mentioned it to have a way to get the body contents in a similar way as in Suave.

src/Server/Auth.fs
alfonsogarciacaro: We've the same code and it's working. I think the problem is [in these lines](https://github.com/SAFE-Stack/SAFE-BookStore/blob/804b10f45ce5566b88f7e241a9ab8e7f4ed2a06e/src/Client/pages/Login.fs#L54-L57), we actually changed them to:

```fsharp
        let! data = response.json()
        return data :?> string
```

Makes sense because the client is asking for a JSON response and the server negotiates (not like Suave that directly sent a text response), so the string comes quoted.

src/Server/FableJson.fs
alfonsogarciacaro: Already too many packages ;) I guess adding a `negotiateWithJson` or similar to Giraffe so you can pass a custom Json serializer would work :+1:

src/Server/Program.fs
alfonsogarciacaro: Actually it'd be nice to have some CLI arguments too. Right now the program gets the config by a mix of CLI args, checking env vars and guessing ;) We have some CLI arguments for host, port, root content and client path with defaults although it's quite primitive. We probably should use Argu :+1:

forki:
@alfonsogarciacaro so content negotiation issue is in client code. Lol. I'm so blind

alfonsogarciacaro:
:+1: For using NETCore config. In the C# ASP.Net Core Web API template they use JSON files for some configuration like the logger or the port. Not sure if it's related. Now we're using `UseUrls` to configure the port for example.

@forki Yes, actually the client is **asking for a JSON** as it adds the `application/json` ContentType header. The Suave version was wrong in a sense because it returned plain text nonetheless. Changing the request header to `plain/text` should also fix the problem I guess.

forki:
Ok I removed the header. But did not work. What's the default?

alfonsogarciacaro:
Giraffe [defaults to JSON response](https://github.com/giraffe-fsharp/Giraffe/blob/60c29e2e8d21886b853066fef2492b8450f12a87/src/Giraffe/HttpHandlers.fs#L393) so if you want `plain/text` I guess you need to specify that.

vasily-kirichenko:
Does this PR hooks the Fable Json serializer so that when we write `json x next ctx`, it's always used?

forki:
No. It does not

Am 16.12.2017 15:01 schrieb "Vasily Kirichenko" <notifications@github.com>:

vasily-kirichenko:
@forki :( For me, it was a real puzzle to fix serialization issues due to usage Fable Json on Fable side and Giraffe's one (vanilla Newtonsoft.Json?) - on the server side.

forki:
Yes I know that's why I think we need to patch that directly into giraffe
or create a new giraffe.fable package that somehow fixes the problem

Am 16.12.2017 15:07 schrieb "Vasily Kirichenko" <notifications@github.com>:

gerardtoconnor:
Giraffe.Fable module/namespace probably makes sense to override JSON and other handlers with preferred Fable behaviour

alfonsogarciacaro:
@gerardtoconnor The only thing that Fable needs from the server side is [passing the Fable.JsonConverter to Newtonsoft.Json](http://fable.io/docs/interacting.html#json-serialization), nothing else. It looks a bit too much for me publishing and maintaining a full Giraffe.Fable package to override all the HttpStatusCode handlers, etc. Would it be possible to make the JSON serializer in Giraffe configurable? This would benefit other users with similar needs as well.

alfonsogarciacaro:
Wrote https://github.com/giraffe-fsharp/Giraffe/issues/178

forki:
@alfonsogarciacaro I want to have a opionated version that works ootb with
fable. And websockets that work nicely with elmish. So eventually it will
happen anyway. Fable is most important Client for me

Am 16.12.2017 18:00 schrieb "Alfonso Garcia-Caro" <notifications@github.com

gerardtoconnor:
It is currently already possible to override/load with custom JsonConveter for both custom json handler and ctx extension. given Fable is so popular for F# users, it might be worth building in a fable specific json handler as adds minimal code but makes fable usage far easier. Built in `fjson` handler, alt namespace, or some other specialised handler with 
JsonConveter built in that is light worth considering. Can open issue to flesh out with @dustinmorris

src/Server/Program.fs
vasily-kirichenko: @alfonsogarciacaro no idea really :) I doubt because it's used in this such an ubiquitous `CreateDefaultBuilder` method.

dustinmoris:
Hey, I hear your issues. I think @alfonsogarciacaro has made a good point that if there would be an easy way to configure json once per app (which would not only benefit Fable but any app which doesn't use vanilla Newtonsoft - so not that uncommon) then it would be a much nicer experience. I have written down a potential solution in the related Giraffe issue. Have a look and let me know!

isaacabraham:
I was chatting with @forki about this the other day. Definitely being able to plug in any Newtonsoft JSON serializer would be the way to go for me - this way all of the excellent functionality in Giraffe like the easy model binding from forms etc. would just magically work.

dustinmoris:
@isaacabraham Cool, well have a look at the [open issue](https://github.com/giraffe-fsharp/Giraffe/issues/178) and maybe let us know if you prefer the custom `JsonSerializerSettings` appraoch or the complete custom `IGiraffeJsonSerializer` approach.

forki:
Ok. It's green. Question is how do we want to proceed vs suave version.

vasily-kirichenko:
@forki as for me, it's not very green :) About Suave - just deprecate it?

forki:
Circle ci is expected to be red. But damn what's going on with Travis?

forki:
It's failing during paket restore? Wtf

alfonsogarciacaro:
Sometimes happens if packages take a big long to download, I just restarted the build :)

forki:
Seems restore is green on master. So something is weird here.

forki:
It

forki:
@matthid I guess we should limit the number of parallel downloads somehow.

forki:
Can someone with Mac or Linux repro locally? Or is it only Travis?

matthid:
@forki can you please give a tldr about the download limit and the problem?

forki:
My Current assumption issue that Travis kills the build because no output
is sent for a while. Paket should buffer package downloads and only have
like 5 at a time.

Am 17.12.2017 12:38 schrieb "Matthias Dittrich" <notifications@github.com>:

forki:
mhm. lol. Instead of blocking everything for > 10min the whole download is now done in 32 seconds.

src/Server/FableJson.fs
forki: https://github.com/giraffe-fsharp/Giraffe/pull/179 this would allow us to patch the negotiation rules easily.

src/Server/Auth.fs
forki: just found   
    member this.ReadBodyFromRequestAsync() =
        let body = this.Request.Body
        use reader = new StreamReader(body, true)
        reader.ReadToEndAsync()

theimowski:
have you guys thought about how Giraffe could be added to [SAFE template](https://github.com/SAFE-Stack/SAFE-template)? Should one have an option to choose from Suave / Giraffe for backend?

isaacabraham:
That has always been my dream - that the SAFE template allows one to choose between eg Suave / Giraffe / Freya etc., and the same for e.g. Azure Storage / Local fs / AWS S3  etc.. Perhaps that's unrealistic but yes, having an option for which one you want would be fantastic

theimowski:
The template engine is bit cumbersome, but hey Christmas is coming, why don't make at least some dreams come true ;)

alfonsogarciacaro:
I haven't worked with the options in the template engine but my gut feeling is it may be complicated and separate templates for each server could be an easier solution. But @MangelMaxime can give a more informed opinion.

About access to Azure Storage / Local fs / AWS S3, etc... I guess these should go in packages (ideally with some common interfaces) instead of putting all of them in the template.

BTW, there's not much to see yet, but I'm working to move most of the boilerplate to build Fable apps into an [independent npm package](https://github.com/fable-compiler/fable-scripts) so it can easily be shared across templates.

cartermp:
I suspect that server backend is going to be the one that 90% of people will want to configure:

* Suave
* Giraffe
* Freya
* NPM-fable-lite (or whatever it's called)

So I think it's worth getting that going and seeing how things go from there.

colinbull:
@vasily-kirichenko Finally, managed to get some time to pull out a first cut of the configuration I was talking about.  

https://github.com/colinbull/StrongNetCoreConfig/tree/master

Pretty rough as I had to do alot of tidying up from what we actually using but it seems to work. Currently two known issues, 

* The first time any config is changed you will get updates for all configuration files that are registered. 
* It currently relies on `IHostingEnvironment` so has a dependency on `aspnetcore`. I'm pretty sure this can be refactored away thou.

dustinmoris:
@alfonsogarciacaro @isaacabraham @cartermp @theimowski @forki 
This thread has inspired me to add 3 different view engine options to the giraffe-template and I wrote a blog post which explains how this can be easily done: https://dusted.codes/extending-the-giraffe-template-with-different-view-engine-options

This way you should be able to easily provide an option between different backends.

theimowski:
FYI I just released SAFE template v0.4.0 with `--Server` parameter to specify either suave or giraffe. Should be available within hour when nuget indexes the package. Thanks @dustinmoris for showcasing choice template options

forki:
ready for final review. Will merge later today

PR #261, opened 2017-12-20
Check version before installing dotnet cli

kunjee17:
I have added version check and also printing if something is wrong. Now it will check and install if and only if that is required.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -110,7 +110,9 @@ Target "Clean" (fun _ ->
 )
 
 Target "InstallDotNetCore" (fun _ ->
-    dotnetExePath <- DotNetCli.InstallDotNetSDK dotnetcliVersion
+    printfn "Cli is installed %A" <| DotNetCli.isInstalled()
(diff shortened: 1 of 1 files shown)


forki:
That is already part of fake. If it doesn't work for you we need to fix it over there

forki:
That is already part of fake. If it doesn't work for you we need to fix it over there

kunjee17:
@forki it doesn't work for me. It was installing dotnet core again and again for me. I ll have a look at it in over Fake.

rommsen:
@kunjee17 we had the same problem but updating fake to the latest version fixed it

kunjee17:
@rommsen thanks man. I ll update it and have a look.

forki:
actually I forgot to document that we are on 2.1.3 already: https://github.com/SAFE-Stack/SAFE-BookStore/commit/5921d6d46aa8ce297a7297ff33e1f467523d58ef

kunjee17:
@forki @rommsen updating Fake did the trick. thanks man...

PR #262, opened 2017-12-22
Use Fable serializer to handle DUs

forki:


forki:
@alfonsogarciacaro somehow I think this file should go somewhere directly into fable.*

alfonsogarciacaro:
As this is a browser specific API, it cannot go into Fable.Core, maybe Fable.PowerPack?

forki:
yes powerpack sounds good

src/Client/Utils.fs
```diff
@@ -1,13 +1,15 @@
...
-let save key (data: 'T) =
-    Browser.localStorage.setItem(key, JS.JSON.stringify data)
+let inline delete key =
+    Browser.localStorage.removeItem(key)
 
-let delete key =
-    Browser.localStorage.removeItem(key)
\ No newline at end of file
+let inline save<'T> key (data: 'T) =
+    delete key
```
alfonsogarciacaro: Why `delete key`? `Browser.localStorage.setItem` will just overwrite if I'm not wrong.

alfonsogarciacaro:
Added in Fable.PowerPack 1.3.2

src/Client/Utils.fs
forki: Yes that was an attempt to fix a bug that I had. Not needed at all.

PR #277, opened 2018-01-23
Make book list a list of components

vbfox:
Instead of directly including the html elements in the parent table :

* An array of elements is exposed to react with a key (Not a really good key but it'll do)
* Each element is a separate component that will appear as such in react dev tools

**Notes:**
* Sorry for the whitespace diff, I have it in auto cleanup in vscode, I can revert
* I'll PR later to use PureComponent instead of a stateless functional one (They aren't as optimized as they can be un-pure) as it give better perf and the immutability of F# record is a perfect match
* I didn't add a key to the `Book` type but I think it should have one as it's a best practice in react anyway (And it seem that the `AzureTable` server side is reduced to using a poor `RowKey` that will create problems) .

### DIFF
--- a/src/Client/pages/WishList.fs
+++ b/src/Client/pages/WishList.fs
@@ -13,7 +13,7 @@ open Fable.PowerPack
 open Fable.PowerPack.Fetch.Fetch_types
 open ServerCode
 
-type Model = 
+type Model =
   { WishList : WishList
     Token : string
     NewBook : Book
@@ -40,7 +40,7 @@ type Msg =
 let getWishList token =
     promise {
         let url = ServerUrls.WishList
-        let props = 
+        let props =
             [ Fetch.requestHeaders [
                 HttpRequestHeaders.Authorization ("Bearer " + token) ]]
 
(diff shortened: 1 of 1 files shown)


src/Client/pages/WishList.fs
```diff
@@ -243,12 +243,27 @@ let newBookForm (model:Model) dispatch =
...
                     ]
-                ]                    
-            ]        
+                ]
+            ]
         ]
     ]
 
-let view (model:Model) (dispatch: Msg -> unit) = 
+type [<Pojo>] BookProps = { key: string; book: Book; removeBook: unit -> unit }
```
forki: @alfonsogarciacaro I wonder if we really need pojo everywhere?

alfonsogarciacaro: For Fable 1.x unfortunately yes. If you don't use it, Fable will create a JS class for the record and React will complain as it expects plain objects.

For Fable 2.0 it won't be needed as we're planning to compile records always as plain objects :)

forki:
I think your react patch arrived on nuget. Can you please send the second part?

vbfox:
Yes i'll do that.

But I have a few questions on what is the exact approach for this project?
* Should I try to add a functional component somewhere else to demonstrate the feature ?
* Should I comment the code explaining WHY things are done this way so that it act as a mini-inline-tutorial ?

forki:
whatever you think is worth tha we spread it through the ecosystem ;-)

Issue #274, opened 2018-01-23
browser refresh fails in wishlist web part


kfrie:
### Description

A browser refresh on the `#wishlist` part when logged in forces the page into a permanent white screen until browser cache is cleared. This does not happen with the [hosted version](http://fable-suave.azurewebsites.net) of the BookStore sample. 

### Repro steps

![reloading_failure](https://user-images.githubusercontent.com/3251459/35290628-adfe0912-0062-11e8-8975-385409ea6ebe.gif)

### Expected behavior

see the same page in the same visual state as before

### Actual behavior

white screen

### Related information 

* Windows 10 Education Version 10.0.16299 Build 16299
* tested with Chrome Version 63.0.3239.132 (Official Build) (64-bit)
* tested with Opera 50.0.2762.58 (PGO)
* repro step shows dev environment, but also happens in production
* commit e17553275a217899ed104802fce48f387f1596f8
* `dotnet --version` = 2.1.3

kfrie:
#### More detailed information

![failure](https://user-images.githubusercontent.com/3251459/35394407-016809e6-01e8-11e8-9c71-853df60db6b8.gif)
![2018-01-25 15_53_08](https://user-images.githubusercontent.com/3251459/35394478-2987cd30-01e8-11e8-8b43-da112f2d3c67.png)

Issue #278, opened 2018-01-25
Failure on first-time build


dsyme:
Steps

- Fork  + clone master a669d99a279c1be112c4e4f9d2f230fd9c7e36c1
- .\build.cmd

Got this:

![image](https://user-images.githubusercontent.com/7204669/35390819-98d62be6-01d4-11e8-81d6-4be9cd1151aa.png)

kfrie:
seems like the UI Tests fail since some UI elements have been changed, but the tests have not been  updated. As a workaround you can simply comment them out in `build.fsx` to get it working:

```fsharp
Target "Build" DoNothing
Target "All" DoNothing

"Clean"
  ==> "InstallDotNetCore"
  ==> "InstallClient"
  ==> "SetReleaseNotes"
  ==> "BuildServer"
  ==> "BuildClient"
  ==> "BuildServerTests"
  ==> "RunServerTests"
  // ==> "BuildClientTests"
  // ==> "RenameDrivers"
  // ==> "RunClientTests"
  ==> "BundleClient"
  ==> "All"
  ==> "CreateDockerImage"
  ==> "TestDockerImage"
  ==> "PrepareRelease"
  ==> "Deploy"
```

kfrie:
I investigated and the UI elements are in sync with the tests. I think something got messed up with user state on the client side. Probably the page gets rendered empty like in #274

dsyme:
BTW switched to do ``build run`` and everything was ok from the first-time user perspective.  I hadn't seen the need for ``run`

forki:
The "build run" is what starts everything in watch mode or what I like to
call "all the time Mode".

Am 25.01.2018 20:07 schrieb "Don Syme" <notifications@github.com>:

kfrie:
yeah and it is an incredible development experience. 

I am also impressed with the deployment and release stuff you get out of the box (this is really important for adoption of commercial use I think). 
Release + Tag on GitHub, push to Docker Registry and deployment to Google Cloud an Azure with 2 commands once everything is set up. Awesome!

PR #279, opened 2018-01-25
add type annotation for loadUser() to prevent deserialize error

kfrie:
this fixes issues #274 #278


forki:
that looks rather important!

savioret:
Thank you so much

Issue #234, opened 2017-11-03
Build project in docker it self


kunjee17:
It is more of a question then an issue. So, feel free to close it. And I m docker noob, so don't mind my queries. Didn't find better place to ask so asking here.

```
# Sample contents of Dockerfile
 # Stage 1
 FROM microsoft/aspnetcore-build AS builder
 WORKDIR /source

 # caches restore result by copying csproj file separately
 COPY *.csproj .
 RUN dotnet restore

 # copies the rest of your code
 COPY . .
 RUN dotnet publish --output /app/ --configuration Release

 # Stage 2
 FROM microsoft/aspnetcore
 WORKDIR /app
 COPY --from=builder /app .
 ENTRYPOINT ["dotnet", "myapp.dll"]
```

This snippet is from official aspnetcore build image. Can't we do similar thing with this project. 

Stage1 to build & test and Stage 2 to run.

Question 
1. It will create two images or single image ?
2. In case of two images what will happen to second image?
3. In case of one image how can it will be two, one will `build and test` and other will `deployed and run`.

Again I apologise if question is totally out of context.

Krzysztof-Cieslak:
Yes, we could do build inside a docker - actually, that was the plan when I've been thinking about CirlceCI build, I need to get back to it.

The problem is that our docker image needs to be a bit more complex than simple ASP.Net Core one - we need to install node (for Fable parts), mono (Paket & FAKE), and .Net Core (well, that is also installed with FAKE script, so can be omitted in the image)

kunjee17:
@Krzysztof-Cieslak I guess then mono image would be better. But anyways thanks for answer. 👍

PR #282, opened 2018-01-28
 Build inside the Dockerfile

vbfox:
Also:
* Bump SDK version
* Fixes `/Images/` vs `/images/` on linux

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+/temp/
+/node_modules/
+/packages/
+/paket-files/
+/bin/
+/.github/
+/.vscode/
+/.fake/
+**/bin/
+**/obj/
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,8 @@
 # Auto detect text files
 * text=auto
 
+*.sh text eol=lf
+
 # Custom for Visual Studio
 *.cs diff=csharp text=auto eol=lf
 *.vb diff=csharp text=auto eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-FROM microsoft/aspnetcore
+FROM microsoft/dotnet:2.1.1-aspnetcore-runtime-alpine3.7
+WORKDIR /bookstore
 COPY /deploy .
-WORKDIR .
 EXPOSE 8085
 ENTRYPOINT ["dotnet", "Server.dll"]
--- a/README.md
+++ b/README.md
@@ -317,6 +317,33 @@ testCase "login with test user" <| fun () ->
 
 [Paket](https://fsprojects.github.io/Paket/) is a dependency manager and allows easier management of the NuGet packages.
 
+## Docker
+
+The project can be executed in [docker](https://www.docker.com/) and can even be built there.
+
+### Running in docker
+
+It is possible to build the project locally and run it in a docker container. While it isn't really useful in itself (It's easier to run it locally without docker) it allow you to test the docker images that will later be [deployed](#Deployment) without needing a full deployment.
+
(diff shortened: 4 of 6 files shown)


forki:
This means everyone who tries it needs docker installed?

vbfox:
What it changes or not:

* Normal build still works locally as before
* Docker build (As in `docker build .`) now run the full tests+build in a container and generate a working container. It can work with just docker installed (No .Net framework, no .net Core, no Fsharp)
* The build done in docker is done on linux whatever the host OS is doesn't count
* First docker build is very slow because it download mono and paket and yarn download half the planet
* As the docker builder download things entreprise proxies are pretty much guaranteed to not work (Especially `apt` as it doesn't respect `HTTP_PROXY` and require custom config.
* Full deploy with docker now does the build twice (Once on the host PC and once inside docker)

A few decisions to make:

* Do we provide 2 Dockerfile one doing the build inside and one outside ?
* If the "build in docker" version is the one used in the FAKE build maybe the tasks in questions should not depend on the local build ?

vbfox:
Also the idea for the feature (I mean except from me wanting to play with docker) is issue #234.

kfrie:
For those people who use Docker and are sceptical towards SAFE and F# ecosystem this is a great thing. It allows them to explore BookStore without cluttering their system. Something I always do if I explore an unknown ground :)

forki:
@kfrie we only need to make sure the other way holds as well - people who don't want to docker need to be able to run it.

kfrie:
I assumed it is served as an option :) But to address the open questions of @vbfox : 

* We must ensure that this Docker thing is just additive, so we can't touch existing `build.fsx` targets and current Dockerfile. So maybe by create a subfolder `DockerizedBuild` where you place the new file
* Then create a "Docker-only" variant of every depending target and define a 2nd, Docker-only build chain in "build.fsx". Bypassing tests could be a build parameter
* Update README.md and present the Docker-only option.

vbfox:
* Moved to a separate dockerfile (Not in a directory dockerfiles hate directories)
* Add some info on usage in readme (For both the old and the new dockerfile)

kfrie:
@vbfox  Good to know, theres always something to learn for a Docker convenience user like me 😅

vbfox:
@kfrie Essentially `docker build` is using inception, it's first packaging the dir passed as parameter in a zip, spinning a special container, sending the zip to the container and building the dockerfile here... so the second parameter is crucial.

if you have your dockerfile in a directory it can work by doing `docker build -f dir/Dockerfile .` (from the parent) or `docker build ..` in the child but it's hard to remember, users get tripped by it and you don't gain anything from placing it along the other ones.

vbfox:
@forki is there a way for this PR to move forward or should I close it ?

forki:
soryy for got about it. can you please rebase on master?

vbfox:
Here is a new versions with quite a few changes:

* I changed the normal Dockerfile to use a more recent aspnet image based on alpine (smaller) & also use a pinned version of the aspnet image
* The image is now based on [vbfox/fable-build](https://hub.docker.com/r/vbfox/fable-build/) because it's a configuration I needed elsewhere & there is no need to copy it
* The docker build no longer run the tests. This is because the client ones can't be run as-is. As the driver moved to chrome it would need a X-Windows subsystem and a full install of chrome to run, it's possible in docker but quite involved (See https://hub.docker.com/r/robcherry/docker-chromedriver/~/dockerfile/ ). So I made the `BundleClient` target in FAKE no longer depend on the tests and use that.

(I might do some sort of `-full` version of `vbfox/fable-build` that can run the tests at some point)

Issue #283, opened 2018-01-29
Build fails on SDK 2.1.300-preview1-008009 at BuildServer step


cartermp:
SDK link [here](https://dotnetfeed.blob.core.windows.net/orchestrated-release-2-1/20180126-01/final/assets/Sdk/2.1.300-preview1-008009/dotnet-sdk-2.1.300-preview1-008009-win-x64.exe).

Info:
```
Windows 10
VS:D15.6/27326.01(15.6 Preview 4) - internal VS build
Node.js: V8.9.4 LTS
Yarn --version: 1.3.2
Paket version: 5.132.4
CLI/SDK: 2.1.300-preview1-008009
```

Failure from running `build.cmd`:
```
System.Exception: dotnet build failed
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1379.Invoke(String message)
   at FSI_0005.Build.runDotnet(String workingDir, String args) in C:\Users\vseqa1\Documents\test\SAFE-BookStore-master\build.fsx:line 64
   at FSI_0005.Build.clo@119-5.Invoke(Unit _arg3) in C:\Users\vseqa1\Documents\test\SAFE-BookStore-master\build.fsx:line 120
  at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in D:\code\fake\src\app\FakeLib\TargetHelper.fs:line 626
```

cartermp:
Note: leaving this here for now so that it's tracked, since we did pick this up when building against this project. We'll route this elsewhere if it turns out not to be an issue with the SAFE project

cartermp:
Log here: 
[FullLog.txt](https://github.com/SAFE-Stack/SAFE-BookStore/files/1676123/FullLog.txt)

Relevant:

```
C:\Users\vseqa1\Documents\SAFE-BookStore-master\src\Server\Auth.fs(13,13): error FS0072: Lookup on object of indeterminate type based on information prior to this program point. A type annotation may be needed prior to this program point to constrain the type of the object. This may allow the lookup to be resolved. [C:\Users\vseqa1\Documents\SAFE-BookStore-master\src\Server\Server.fsproj]
[...]

PR #284, opened 2018-01-31
Try to repro #283

forki:
#283


forki:
Closing in favor of #293 which repros on newer version

PR #285, opened 2018-02-01
minor formatting updates

dsyme:
I was showing this code to a Xamarin team member today and noticed a few places where comments or some textual cleanup might help in readability


dsyme:
I have compiled as far as ``BuildClient`` but had trouble running tests on OSX as I get ``node not found`` even though I have it installed.  I haven't traced into that

PR #286, opened 2018-02-03
Use Saturn

Krzysztof-Cieslak:
Feedback welcomed.


forki:
Very very nice

src/Server/WishList.fs
```diff
@@ -2,35 +2,37 @@
...
 open ServerErrors
 open ServerCode.Domain
 open Microsoft.AspNetCore.Http
 open Microsoft.Extensions.Logging
 open System.Threading.Tasks
+open System.Security.Claims
 
 /// Handle the GET on /api/wishlist
 let getWishList (getWishListFromDB: string -> Task<WishList>) next (ctx: HttpContext) =
-    Auth.useToken next ctx (fun token -> task {
```
isaacabraham: @Krzysztof-Cieslak From an educational point of view, what does the `Auth.useToken` get replaced with in Saturn?

src/Server/Auth.fs
```diff
@@ -10,31 +10,14 @@ let login next (ctx: HttpContext) = task {
...
-        | Some t when t.StartsWith "Bearer " ->
-            let jwt = t.Replace("Bearer ","")
-            match JsonWebToken.isValid jwt with
-            | None -> return! FORBIDDEN "Accessing this API is not allowed" next ctx
-            | Some token -> return! f token
-        | _ ->
-           return! BAD_REQUEST "Request doesn't contain a JSON Web Token" next ctx
-    | _ -> 
-        return! BAD_REQUEST "Request doesn't contain a JSON Web Token" next ctx
-}
```
isaacabraham: @Krzysztof-Cieslak I assume that this whole block is rendered obsolete by Saturn?

src/Server/Program.fs
```diff
@@ -65,16 +44,16 @@ let main args =
...
-            .UseWebRoot(clientPath)
-            .UseContentRoot(clientPath)
-            .ConfigureLogging(configureLogging)
-            .Configure(Action<IApplicationBuilder> (configureApp database clientPath))
-            .UseUrls("http://0.0.0.0:" + port.ToString() + "/")
-            .Build()
-            .Run()
+
+        let app = application {
+            router (WebServer.webApp database)
```
isaacabraham: Is it worth also replacing / simplifying the `webApp` with Saturn as well?

src/Server/Program.fs
```diff
@@ -65,16 +44,16 @@ let main args =
...
+            url ("http://0.0.0.0:" + port.ToString() + "/")
+
+            use_jwt_authentication JsonWebToken.secret JsonWebToken.issuer
+            use_static clientPath
+            use_gzip
+        }
+        run app
```
isaacabraham: @Krzysztof-Cieslak I can see most of the mappings from before -> after, but do the calls for e.g. `CreateDefaultBuilder()` etc. simply get put in for free within the `application { }` CE?

src/Server/Program.fs
```diff
@@ -3,55 +3,34 @@ module ServerCode.Program
...
-       .UseStaticFiles()
-       .UseGiraffe (WebServer.webApp db root)
-       
-let configureLogging (loggerBuilder : ILoggingBuilder) =
-    loggerBuilder.AddFilter(fun lvl -> lvl.Equals LogLevel.Error)
-                 .AddConsole()
-                 .AddDebug() |> ignore
-
-
-
```
isaacabraham: @Krzysztof-Cieslak how / where is this replaced with by Saturn?

src/Server/Auth.fs
Krzysztof-Cieslak: Yes, indeed. Saturn default JWT helpers are using authorization implementation provided by ASP.NET which means that all handling of cases where token is not provided, token is not valid etc is already provided. 

It changes a bit a way how you use token - in old implementation `useToken` was "higher order handler" which fed token to out actions. In the new one, ASP.NET is populating `HttpContext.User` field. While it includes some ASP.NET magic it makes the authentication system totally pluggable - you can easily change authentication to Cookies, or external OAuth or anything else that's supported by ASP.NET and framework will still populate same `HttpContext.User` field with authentication information.

src/Server/Program.fs
Krzysztof-Cieslak: * `UseGiraffe (WebServer.webApp db root)` - is replaced by `router` call in `application` CE
* `UseGiraffeErrorHandler(errorHandler)` - is provided by default by Saturn, can be overriden by using `error_handler` in `application` CE
* `UseStaticFiles` - is replaced by `use_static in `application` CE
* logger configuaration - hmm... good question, I think I haven't added it yet to Saturn. Will put on my TODO list.

src/Server/Program.fs
Krzysztof-Cieslak: Yes, Satur tries to hide some of this additional fluff away - so `CreateDefaultBuilder` and `Build` are provided automatically by Saturn. `UseWebRoot` and `UseContentRoot` are set inside `use_static` helper. 

And `run app` function is just `app.Run()` call

src/Server/WishList.fs
Krzysztof-Cieslak: Checking if the request is authenticated is done by `pipe_through jwtAuthentication` in `secured` router in WebServer.fs file.

As I've mentioned in the earlier comment this is using ASP.NET functionality and `Challenge` functionality which checks if the request is authorized and provides appropriate, standard responses if the request is not authorized. If the request is authorized we just go to the handler for a particular path (`getWishlist` for example) , and here we can use `ctx.User` to get information about current user - this field is automatically populated by ASP.NET

src/Server/Auth.fs
Krzysztof-Cieslak: So it's not really made obsolete by Saturn - it's just side effect of Saturn using ASP.NET mechanisms for authentication helpers. 

Same would happen if we moved to ASP.NET authentication without using Saturn, and just Giraffe's `challenge` and `requiresAuthentication` handlers

src/Server/WishList.fs
Krzysztof-Cieslak: Again, it's just side effect of Saturn using ASP.NET mechanisms for authentication helpers.

Same would happen if we moved to ASP.NET authentication without using Saturn, and just Giraffe's challenge and requiresAuthentication handlers

Krzysztof-Cieslak:
In general, I believe that Saturn advantages are shown better on bigger, more complex applications (for example I don't use `controller` abstraction at all in this PR - it just doesn't make sense for service with 3 endpoints but is designed to be used in a more REST-ish type of applications).

But I also think it nicely shows some of the core ideas behind Saturn such as hiding configuration boilerplate, providing opinionated helpers (for example authentication in this case)

alfonsogarciacaro:
Would it be possible to remove (most of) `try .. with` expressions? If we're using a higher level framework I have the feeling it should manage logging and error reporting for you.

Ideally it should also manage content negotiation and data serialization, so `getWishList` can return directly the wish list (maybe using a transform function to change the type) instead of explicitly doing `FableJson.serialize wishList next ctx` (you would have configured the JSON serializer when starting the app).

BTW, should WishList be a controller to show users how controllers work in Saturn?

Krzysztof-Cieslak:
> Would it be possible to remove (most of) try .. with expressions? If we're using a higher level framework I have the feeling it should manage logging and error reporting for you.

I'm not 100% sure about that, we don't want to put abstraction layer too high. 

> Ideally it should also manage content negotiation and data serialization, so getWishList can return directly the wish list (maybe using a transform function to change the type) instead of explicitly doing FableJson.serialize wishList next ctx (you would have configured the JSON serializer when starting the app).

In general yes, I totally agree and hope to do something like that. It's just not there yet, I need to investigate what's a good way of configuring serializer/deserializer.

> BTW, should WishList be a controller to show users how controllers work in Saturn?

I kinda feel that for such simple sample it would be actually more noise... I'm planning to do bigger S(aturn)AFE sample app showing more complex things, architecture I want to suggest, etc. 

Also, I feel fact that controller is not here nicely show that Saturn is not all or nothing type of framework but you can use to progressively update existing Giraffe application.

dustinmoris:
> In general yes, I totally agree and hope to do something like that. It's just not there yet, I need to investigate what's a good way of configuring serializer/deserializer.

This is literally coming in Giraffe this week! I am prepping a new release since Monday... was writing a lot of new documentation. As soon as I am ready I'll hit the publish button.

Krzysztof-Cieslak:
@dustinmoris, great news ❤️

tforkmann:
Whats the status of that pr?
Would like to see how to use Saturn in a SAFE project correctly.

Thanks!

Krzysztof-Cieslak:
I'll try to rebase and port to latest version today / tomorrow

alfonsogarciacaro:
Ping!

forki:
Yay. Thanks so much.

PR #287, opened 2018-02-06
Change bootstrap.css to explicit 3.3.7

AkosLukacs:
Because the latest Bootstrap on CDN now is 4.0+, broke couple of things...


forki:
yes! very important fix! thank you

PR #289, opened 2018-02-07
Use a free port for fable daemon

vbfox:
Avoid things breaking up when other projects are building or being watched at the same time

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -138,7 +138,7 @@ Target "InstallClient" (fun _ ->
 
 Target "BuildClient" (fun _ ->
     runDotnet clientPath "restore"
-    runDotnet clientPath "fable webpack -- -p"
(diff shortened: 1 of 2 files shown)


build.fsx
```diff
@@ -138,7 +138,7 @@ Target "InstallClient" (fun _ ->
 
 Target "BuildClient" (fun _ ->
     runDotnet clientPath "restore"
-    runDotnet clientPath "fable webpack -- -p"
+    runDotnet clientPath "fable webpack --port free -- -p"
```
forki: This is not for watch mode. I don't think it's needed here. Right?

vbfox:
It's actually where I need it. I had an other VSCode watching some other project & it stopped this one from building.

Don't understand why it isn't the default.

I also disable the task that kill dotnet in this build.fsx, with more and more things using dotnet it's not so great...

PR #290, opened 2018-02-10
Updated Giraffe

dustinmoris:
Updated Giraffe to 1.0.0.

Sorry for the removal of whitespace on line endings, this is something which my VS Code does automatically. If you want me to exclude it I can change the PR again.

I have also not made any changes to configure JSON yet, because I wanted to show that the upgrade from Giraffe beta to 1.0.0 is actually very minimal if you don't make use of any new features yet.


forki:
looks good. thanks

PR #291, opened 2018-02-10
Removed custom JSON handler

dustinmoris:
Configured Giraffe's JSON serializer to use the FableConverter so that all JSON (de-)serialize methods use the Fable settings. As a result I could remove the custom JSON handlers for binding a model and writing a response.

### DIFF
--- a/src/Server/Auth.fs
+++ b/src/Server/Auth.fs
@@ -7,17 +7,17 @@ open Microsoft.AspNetCore.Http
 
 /// Login web part that authenticates a user and returns a token in the HTTP body.
 let login next (ctx: HttpContext) = task {
-    let! login = FableJson.getJsonFromCtx<Domain.Login> ctx
+    let! login = ctx.BindJsonAsync<Domain.Login>()
 
     try
-        if (login.UserName <> "test" || login.Password <> "test") && 
+        if (login.UserName <> "test" || login.Password <> "test") &&
            (login.UserName <> "test2" || login.Password <> "test2") then
             return! failwithf "Could not authenticate %s" login.UserName
         let user : ServerTypes.UserRights = { UserName = login.UserName }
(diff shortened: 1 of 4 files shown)


src/Server/Auth.fs
```diff
@@ -7,17 +7,17 @@ open Microsoft.AspNetCore.Http
...
 
     try
-        if (login.UserName <> "test" || login.Password <> "test") && 
+        if (login.UserName <> "test" || login.Password <> "test") &&
            (login.UserName <> "test2" || login.Password <> "test2") then
             return! failwithf "Could not authenticate %s" login.UserName
         let user : ServerTypes.UserRights = { UserName = login.UserName }
         let userData :Domain.UserData = { UserName = login.UserName; Token = JsonWebToken.encode user }
-        return! FableJson.serialize userData next ctx
+        return! ctx.WriteJsonAsync userData
```
forki: what is happening with next here?

forki:
Hugs

src/Server/Auth.fs
dustinmoris: The response writers don't need a `next` because it wouldn't get used anyway.

When you look at the implementation, you will see that all response writers essentially boil down to `WriteBytesAsync`, which doesn't need a next because it [short circuits the request with `Some this`](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/ResponseWriters.fs#L26) after writing to the body of the response. This is intentional because after writing to the response stream of the `HttpContext` object you cannot make any changes to `HttpContext.Response` anymore otherwise ASP.NET Core will throw an exception that it cannot modify the response after sending it back to the client.

This is also the reason why the [`>=>`/`compose` checks for `ctx.Response.HasStarted`](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Core.fs#L201).

forki: Makes sense. Thanks

PR #292, opened 2018-02-10
Little Refactor

dustinmoris:
Simplified the Auth and WishList modules.

- Removed try-with in WishList functions by moving the error code to the error handler
- Removed try-with from login and return an error response straight away without throwing an exception first
- Changed the Auth useToken to requiresToken which is a handler now which validates the token and then passes the token to another http handler. This makes it possible to chain `Auth.requiresJwtToken` in the router so it is explicitly visible to any user at first sight which routes require authentication.

I have made a few changes which I think make the code a bit tidier and simpler, but it is very subjective, so please reject this PR or feel free to modify if you don't like it!

### DIFF
--- a/src/Server/WishList.fs
+++ b/src/Server/WishList.fs
@@ -1,47 +1,42 @@
 /// Wish list API web parts and data access functions.
 module ServerCode.WishList
 
-open Giraffe
-open RequestErrors
-open ServerErrors
-open ServerCode.Domain
-open Microsoft.AspNetCore.Http
-open Microsoft.Extensions.Logging
 open System.Threading.Tasks
+open Microsoft.AspNetCore.Http
+open Giraffe
+open ServerCode.Domain
+open ServerTypes
 
 /// Handle the GET on /api/wishlist
-let getWishList (getWishListFromDB: string -> Task<WishList>) next (ctx: HttpContext) =
-    Auth.useToken next ctx (fun token -> task {
-        try
+let getWishList (getWishListFromDB : string -> Task<WishList>) (token : UserRights) : HttpHandler =
+     fun (next : HttpFunc) (ctx : HttpContext) ->
+        task {
             let! wishList = getWishListFromDB token.UserName
             return! ctx.WriteJsonAsync wishList
-        with exn ->
-            let msg = "Database not available"
-            let logger = ctx.GetLogger "wishlist"
-            logger.LogError (EventId(), exn, msg)
-            return! SERVICE_UNAVAILABLE "Database not available" next ctx
-    })
+        }
+
+let private invalidWishList =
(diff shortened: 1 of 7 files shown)


dustinmoris:
Ops, sorry need to fix the build errors now. I know why this happened.

Krzysztof-Cieslak:
Copy pasted from Slack ;)

 I don't think having one global error handler is a good idea. I think having it per Saturn's controller would be fine, but not sure about global one.

I think it could work if we talk about "handle globally some generic failures, but handle bounded context specific ones in particular action handler"  it would be fine...but: 
1. this sample is not showing that 
2. people will abuse this single error handler and will put everything there and just from my experience something like that won't work when you scale size of application

It's fine for small microservices type of apps - I'm just not sure if we should promote something that will hurt people when they try to do some more complex apps

forki:
@Krzysztof-Cieslak how about we merge it and then you rebase the saturn one on top of it?

src/Server/WishList.fs
```diff
@@ -1,52 +1,47 @@
...
             let! wishList = ctx.BindJsonAsync<Domain.WishList>()
 
-            if token.UserName <> wishList.UserName then
-                return! FORBIDDEN (sprintf "WishList is not matching user %s" token.UserName) next ctx
-            else
-                if Validation.verifyWishList wishList then
+            match token.UserName.Equals wishList.UserName with
+            | true ->
+                match Validation.verifyWishList wishList with
+                | true ->
```
forki: pattern matching with true and false is a bit unidiomatic in F#

src/Server/WishList.fs
```diff
@@ -1,52 +1,47 @@
...
+let postWishList (saveWishListToDB: WishList -> Task<unit>) (token : UserRights) : HttpHandler =
+    fun (next : HttpFunc) (ctx : HttpContext) ->
+        task {
             let! wishList = ctx.BindJsonAsync<Domain.WishList>()
 
-            if token.UserName <> wishList.UserName then
-                return! FORBIDDEN (sprintf "WishList is not matching user %s" token.UserName) next ctx
-            else
-                if Validation.verifyWishList wishList then
+            match token.UserName.Equals wishList.UserName with
```
forki: using Equals instead of = is usally considered unidiomatic in F#

src/Server/WishList.fs
dustinmoris: Thank you! I didn't know that!

src/Server/WishList.fs
dustinmoris: Ok cool... didn't know that either... I think this is still something I'm used from C#. Should probably read some F# coding standard guidelines somewhere to make myself more familiar with this stuff!

PR #295, opened 2018-02-14
Use localhost

forki:


forki:
@shanselman @davidfowl sorry to bother you about this, but this "SAFE-BookStore" is now something like the canonical sample for "how to web in F#".

As @cartermp reported Microsoft's internal testing of this project revealed that asp.net core vnext is changing behaviour around `.UseUrls("http://.../")` (at least that's what we assume right now).
The issue is that bookstore like many other asp.net apps need to configure the port. 

First answer on stackoverflow gices you: https://stackoverflow.com/questions/37365277/how-to-specify-the-port-an-asp-net-core-application-is-hosted-on

So ` .UseUrls("http://localhost:8085/")` seems the way to do that - but unfortunately this doesn't work inside a docker container. Anoter answer on stackoverflow (see https://stackoverflow.com/a/46950637/145701) suggested to use ` .UseUrls("http://0.0.0.0:8085/")` instead. Which we did for quite some time now.

But according to Microsoft's internal testing with the next version of asp.net core this doesn't make the service available on `http://localhost:8085` anymore - instead you need to explicitly write `http://0.0.0.0:8085` in the browser.

Please note: I'm not fully sure if I report this 100% correctly since I don't had access to that repro. But the thing that I need for sure is: Changing the config from ` .UseUrls("http://0.0.0.0:8085/")` to ` .UseUrls("http://localhost:8085/")` like I did here in the PR breaks the production app inside the docker container in my azure app service.

I would appreciate guidance on how we should fix it - this will probably copied by most SAFE apps.

cartermp:
Just to clarify - we run SAFE bookstore through our manual validation process for .NET Core.

For .NET Core 2.1 Preivew 1 validation, we recently got a bug report that the project ran successfully, but started listening on `0.0.0.0` rather than `localhost`. The code in this project, which binds to `0.0.0.0` via `UseUrls` has been this way for a long time. It was only recently that we noticed it listening on `0.0.0.0` instead of `localhost`.

We determined that this wasn't a ship-blocker for .NET Core 2.1 Preview 1, but the change in behavior means that _something_ was going on differently under the covers, perhaps in ASP.NET Core.

The linked (and quite old) SO question suggests that `UseUrls`and binding to `0.0.0.0` should be used if a project is to be run in a Docker container. Is that good guidance @davidfowl?

davidfowl:
Don't change the code in the application, provide the urls via an env variable when using docker. This is what the aspnetcore docker image does by default:

https://github.com/aspnet/aspnet-docker/blob/f85c35c8d2be2170fe727cf65c139e7b41226f7d/2.0/jessie/runtime/Dockerfile#L4

forki:
Ok thanks, but what does this mean for watch mode? We run the aspnet site
on 8085 during dev in watch mode as well. And that's what apperently breaks
in the new version. Is this change in behavior expected?

Am 15.02.2018 17:43 schrieb "David Fowler" <notifications@github.com>:

isaacabraham:
Why does it matter whether the URL comes from config rather than within the application?

davidfowl:
If you know in *all* deployment targets what the port will be then you can hardcode all you want. That happens to not be in the case in most hosting environments, the environment usually dictates what port you should be listening on.

forki:
@davidfowl yes I do understand that and I thik it's a fair point, but that does not answer the initial question. Does `. UseUrls` behaviour change in asp.net vnext?

davidfowl:
Nothing has changed with UseUrls.

forki:
Ok thank you @davidfowl.

@cartermp So we are back to step zero

Am 16.02.2018 07:42 schrieb "David Fowler" <notifications@github.com>:

forki:
I have to revert this because it breaks docker deployment. I'm really lost here.

PR #306, opened 2018-03-04
Add Server-Side Rendering (SSR)

zaaack:
## **SSR** from [fable-react](https://github.com/fable-compiler/fable-react/blob/master/docs/server-side-rendering.md)!

This PR is actually working, however, it is still waiting fable-react been published, and  [elmish-react's `withReactHydrate `](https://github.com/fable-elmish/react/pull/19) to be merged and published. Currently, it's using paket's GitHub dependences feature.

What's more, here are some breaking changes to add SSR:

1. Urls are `path` not `hash` anymore, which means we would create more routes for each page. Also, click on a direct links would cause a browser refresh which not in the old `hash` urls, I added a [helper function](https://github.com/SAFE-Stack/SAFE-BookStore/compare/master...zaaack:ssr?expand=1#diff-23e919ce464389c07bfc775179e3926fR12) to execute `Elmish.Browser.Navigation.Navigation.newUrl` directly to avoid it, but not sure it's right. @MangelMaxime 

2. Add jwt to cookies to make SSR support rendering login state, but I didn't remove the jwt in localStorage, this might cause problems...

3. Because the Server project is now depending on the Client project's files, edit view files would trigger the server to rebuild which takes too long. Perhaps we can optimize the development process, e.g. adding a fake target which only builds Server project when files in the Server folder changes?


zaaack:
Looks I  forgot the ui test, I'll fix it later...

MangelMaxime:
Seems like the helper is needed, other users mention it. This is because of how browser work, the solution used by react, is to create a custom "link" element.

Idea being, the helper is included by default in the custom "link" so usage is transparent.

forki:
can you please change it to the nuget package?

zaaack:
@forki Just planning to do it. 😉

zaaack:
@forki Updated to nuget packages. 

I also tried some configuration in [dotnet-watch](https://github.com/aspnet/DotNetTools/blob/dev/src/dotnet-watch/README.md) to ignore client files in develop running, but have no luck:
```xml
<?xml version="1.0" encoding="utf-8"?>
<Project>
  <PropertyGroup>
    <IsWatch Condition="'$(Configuration)' == 'DebugSSR'">false</IsWatch>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="../Client/ReleaseNotes.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/Pages.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/Style.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/views/Menu.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/pages/Home.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/pages/WishList.fs" Watch="$(IsWatch)" />
    <Compile Include="../Client/pages/Login.fs" />
    <Compile Include="../Client/Shared.fs" Watch="$(IsWatch)" />
  </ItemGroup>
</Project>
```

build.fsx
```diff
@@ -216,6 +217,30 @@ Target "Run" (fun _ ->
 )
 
 
+Target "RunSSR" (fun _ ->
+    runDotnet clientPath "restore"
```
forki: why do we need this? what is it doing?

forki:
can you please rebase/merge with master

build.fsx
zaaack: This will set a compile directive `DEBUG_SSR` to use webpack-dev-server's bundle file path in the server-rendered path. Ideally, in `RunSSR` we can trigger client and server to rebuild when we edit any file, so we can dev and debug with SSR; but in normal `Run` we would only trigger the server to rebuild when we edit back-end files, and we can dev with webpack-dev-server and use the server code as an API service.

But I cannot pass the `configuration` to dotnet-watch to make it work in different command, currently, we totally ignored client files' change when `watch run` the server.

https://github.com/SAFE-Stack/SAFE-BookStore/pull/306/files#diff-694c6c636cd70bcc23aa716434744995R12
https://github.com/SAFE-Stack/SAFE-BookStore/pull/306/files#diff-3b3c3fd8d4441c71716037ce14cb42b0R14

zaaack:
@forki Done.

paket.dependencies
```diff
@@ -13,10 +13,12 @@ nuget System.Net.NetworkInformation
 nuget jose-jwt
 
 nuget Fable.Core
-nuget Fable.Elmish.Browser
-nuget Fable.Elmish.Debugger
-nuget Fable.Elmish.React
-nuget Fable.Elmish.HMR
+nuget Fable.React 3.0.0-alpha-002
```
forki: @alfonsogarciacaro any timeline on when these things get released?

alfonsogarciacaro: The Fable.React 3 _stablish_ version? At the end we're not going to touch more the API so I guess we can do it now ([this PR](https://github.com/fable-compiler/fable-react/pull/63#issuecomment-370573808) is pending though). I'll have to make some minor but potentially breaking changes for Fable 2, but we can bump another major version then.

MangelMaxime: For the PR you linked, I think a new nuget should be created.

Bumping the major version for Fable 2, is ok no need to hold the prerelease for it I think.

forki:
nuget Fable.Elmish 2.0.0-beta-2
nuget Fable.Elmish.React 2.0.0-beta-2
nuget Fable.Elmish.Browser 2.0.0-beta-1
nuget Fable.Elmish.Debugger 2.0.0-beta-1
nuget Fable.Elmish.HMR 2.0.0-beta-1

are these things released properly?

forki:
@zaaack what's that package-lock.json file?

zaaack:
@forki I'm afraid not, these prerelease versions are added to using this PR: https://github.com/fable-elmish/react/pull/19
 although we can simple add a custom elmish render extension method with out these.

> what's that package-lock.json file?

Probably is added when I ran UITest locally...

forki:
@zaaack yes I'd rather not depend on prereleases for public sample

MangelMaxime:
@zaaack The `package-lock.json` is in general generared by npm5 command.

Perhaps there is a call somewhere to `npm` instead of `yarn` commands.

@forki I am currently porting my work app to the latest prerelease version. Problem, is if we want people to understand how to use this new version we need some more documentation.

@vbfox Should be working on a third article in it Fable + React series, explaining how to write the view functions with Elmish to benefit of react optimization. 

IMO, we need to address this lack of documenation before releasing a stable version.

forki:
@MangelMaxime I'm very much in favor of extending docs - but if we don't break stuff then we should still release the versions and extend the samples. It's always bad to have samples on prereleases

forki:
@zaaack this PR does not run SSR on normal build or does it? what hapens if I do just "build.cmd"?

MangelMaxime:
@forki We are breaking stuff :)

The major breaking change being: arguments order has been reversed in view and update function

forki:
yes just saw that. how about minor then? with the hydrate stuff?

MangelMaxime:
@forki It's probably possible, wait for @et1975 answer.

I will try to update Fulma-Demo with support for the new API and good practice stuff. So we will be able to write the documentation needed for the stable release.

zaaack:
@forki  Just did some refactor for RunSSR, sorry that I still cann't pass some param to the project conf, but I think the current one is more reasonable. 

`./build.sh run` would open webpack-dev-server by default, server-side rendering is working but you won't know unless you open the server's url and build the client bundle.js manually.

`./build.sh runssr` would open server's url, and run `dotnet fable webpack -- -w` to watch build client code, you need to manually refresh your page if you edit your client code, but in this case, you can debug server-side rendering easier.

Neither of these would auto restart your server when you edit your client files.

forki:
well I'm not talking about dev mode. I'm talking about the "build.cmd Deploy" 
What would that use SSR if I deploy to azure?

zaaack:
@forki  I think you need to do nothing, just like before. The client files are already included in server's project conf and built with it,  so it should work automatically.

src/Server/Templates.fs
```diff
@@ -0,0 +1,41 @@
...
+let index (model: Model) =
+  let jsonState = toJson (toJson model)
+  let htmlStr = Client.Shared.view model ignore |> renderToString
+  html []
+    [ head [] [ meta [ _httpEquiv "Content-Type"; _content "text/html"; _charset "utf-8" ] ]
+      title [] [ rawText "SAFE-Stack sample" ]
+      link
+        [ _rel "stylesheet"
+          _href "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css";
+          attr "integrity" "sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u"
```
forki: what's this thing?

src/Server/WebServer.fs
```diff
@@ -16,13 +17,16 @@ let webApp databaseType root =
 
     router notfound [
         GET [
-            route "/" (htmlFile (System.IO.Path.Combine(root,"index.html")))
-            route ServerUrls.WishList => Auth.requiresJwtToken (WishList.getWishList db.LoadWishList)
-            route ServerUrls.ResetTime (WishList.getResetTime db.GetLastResetTime)
+            route PageUrls.Home => Auth.addUserDataForPage Pages.home
```
forki: can you please remove the => and just use ()

src/Server/Templates.fs
zaaack: It's a html5 attribute to verify the subresource, see: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

The `Template.fs` is simply transformed from the `index.html` file: https://github.com/SAFE-Stack/SAFE-BookStore/blob/master/src/Client/index.html#L6

src/Server/Templates.fs
```diff
@@ -0,0 +1,41 @@
...
+open Giraffe
+open ServerCode.Domain
+open ServerCode.FableJson
+open ServerTypes
+open Client.Shared
+open Giraffe.GiraffeViewEngine
+open Fable.Helpers.ReactServer
+
+let index (model: Model) =
+  let jsonState = toJson (toJson model)
```
forki: toJson^2 ?

zaaack: Yeah, I explained this in [the SSR tutorial](https://github.com/fable-compiler/fable-react/blob/master/docs/server-side-rendering.md#step-3-create-your-initial-state-on-the-server-side), I can add more comments here if you think it's needed.

src/Client/index.html
```diff
@@ -6,7 +6,7 @@
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap-theme.min.css" integrity="sha384-rHyoN1iRsVXV4nD0JutlnGaslCJuC7uwjduW9SVrLvRYooPp2bWYgmgJQIXwl/Sp" crossorigin="anonymous">
 		<link rel="stylesheet" href="css/site.css">
-		<link rel="shortcut icon" type="image/png" href="/Images/safe_favicon.png"/>
+		<link rel="shortcut icon" type="image/png" href="/images/safe_favicon.png"/>
```
forki: this wil break the image

src/Client/webpack.config.js
```diff
@@ -42,7 +42,8 @@ module.exports = {
       }
     },
     hot: true,
-    inline: true
+    inline: true,
+    historyApiFallback: true,
```
forki: what#s that doing?

et1975:
@forki I'm still waiting for some figures from @vbfox  that show the breaking changes were worth it. I'd suggest backporting SSR support on v1 branch.

forki:
@zaaack can we make it work without cookies?

src/Client/index.html
zaaack: The images folder is in lower case, I'm afraid upper case won't work in case-sensitive systems, like linux.

src/Client/webpack.config.js
zaaack: This is for url router, with  this webpack-dev-server will route all 404 urls back to index.html. So we can use html5 history api router in the client. See: https://webpack.js.org/configuration/dev-server/#devserver-historyapifallback

zaaack:
@forki Yeah, but without cookies, the server-side rendering cannot contain login state, this means some pages/elements that contain login state shouldn't be rendered in the server-side, and the client still need to fetch APIs after page load, including wishlist, login page, and even home page's header.

src/Client/index.html
forki: https://github.com/SAFE-Stack/SAFE-BookStore/blob/master/build.fsx#L267

zaaack: Got, I guess we should either keep all images folder name as capitalized or lower-cased, I think I hit this problem when I run SAFE-BookStore locally on a Linux laptop.

forki:
@zaaack any chance you can port hydrate down to current elmish?

zaaack:
@forki @et1975  Just send a PR to elmish-react v1.x. https://github.com/fable-elmish/react/pull/21

src/Client/index.html
```diff
@@ -6,7 +6,7 @@
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
 		<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap-theme.min.css" integrity="sha384-rHyoN1iRsVXV4nD0JutlnGaslCJuC7uwjduW9SVrLvRYooPp2bWYgmgJQIXwl/Sp" crossorigin="anonymous">
 		<link rel="stylesheet" href="css/site.css">
-		<link rel="shortcut icon" type="image/png" href="/images/safe_favicon.png"/>
+		<link rel="shortcut icon" type="image/png" href="/Images/safe_favicon.png"/>
```
forki: the same is in the server side rendered version

forki:
something broke on travis

zaaack:
@forki Looks like some error occurred when executing `yarn install`, can you rerun travis ci?

forki:
all green now.

src/Client/pages/Login.fs
```diff
@@ -41,22 +41,23 @@ let authUser (login:Login) =
 
         let body = toJson login
 
-        let props = 
+        let props =
             [ RequestProperties.Method HttpMethod.POST
+              RequestProperties.Credentials RequestCredentials.Sameorigin
```
forki: what's this doing?

zaaack: This would add cookies for `fetch` API,  see: https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch
image

But I guess we don't need it here, I'll remove it.

src/Server/WebServer.fs
```diff
@@ -16,13 +17,16 @@ let webApp databaseType root =
 
     router notfound [
         GET [
-            route "/" (htmlFile (System.IO.Path.Combine(root,"index.html")))
```
forki: I assum index.html can now be deleted completely!?

zaaack: Not really, since we still use the client router, we can customize 404 pages on the client side.

src/Client/App.fs
```diff
@@ -17,13 +17,16 @@ open ServerCode.Domain
 JsInterop.importSideEffects "whatwg-fetch"
 JsInterop.importSideEffects "babel-polyfill"
 
+let handleNotFount (model: Model) =
```
forki: found with d ;-)

forki:
is it possible to start a command in SSR?

forki:
@zaaack in my app I have to kick of a bunch of actions after the page is served. Like loading a map and showing current pos

zaaack:
@forki Yes, you can start any command you like, they would be ignored in SSR. 

And here also have `isomorphicExec` and `isomorphicView` to exec different code or rendering different view between client-side  rendering and server-side rendering in
 Fable.Helpers.Isomorphic. You can use this to easily render a loading component in SSR and render the real component in the client .

zaaack:
I'm assuming the map component is written in js native, and imported by FFI, these native component cannot be rendered in SSR. Also including fetch api based on FFI, unless you are using some isomorphic http client library like this:https://github.com/haf/Http.fs/issues/142 If this land I think we can figure out a way to run some initial fetch command in SSR.

zaaack:
Looks Travis ci need a rerun again.

forki:
@zaaack I don't want them to be rendered server side. I just want that command to spin off after the client loaded the bundle.

zaaack:
@forki Just add commands in your `init` function, they will be ignored in SSR, but executed in the client.

forki:
oh crazy. that works!

forki:
I think I'm ready to take it in. what are others saying?

forki:
still hoping we can find a way without cookies. It's such a hard thing in the EU

zaaack:
@forki This really depends on your needs. If you don't use cookies to store the login state, then the login state won't be auto carried when requesting pages, so SSR won't get the login state and render it. As far as I know, the most use case of SSR is SEO, pages need login are actually not much need to be rendered on the server side.

For SAFE-BookStore, there only remains the home page's SSR makes sense if we don't use cookies, I can remove login/wishList's SSR if you want.

forki:
tbh I think that's what we should do. restrict it to when it's logged out.
It's still super valuable

theimowski:
This look really great. Wondering whether it should be by default or opt-in in SAFE-Template?

forki:
merged! thank you so much @zaaack !

forki:
https://twitter.com/sforkmann/status/976736877586903042

Issue #313, opened 2018-03-07
Passing parameters to webpack and webpack-dev-server


cgravill:
### Description

With the switch to Webpack 4.x there's a warning I wanted to address:

```
WARNING in configuration
The 'mode' option has not been set. Set 'mode' option to 'development' or 'production' to enable defaults for this environment.
```

I've done this in other projects by adjusting the packages.json scripts to pass in `--mode development` then invoke the webpack server via `dotnet fable yarn-start`

In this sample, because it uses the commands `fable webpack` or fable `webpack-dev-server` they're explicitly matched:

https://github.com/fable-compiler/Fable/blob/ca91c7962edbcff964132d138954ce27c2ce7cc0/src/dotnet/dotnet-fable/Main.fs#L255

and most additional arguments are dropped (the port argument is explicitly read and passed).

### Repro steps

run `build run` and wait for the warning

### Expected behavior

Be able to configure webpack via build.fsx.

### Actual behavior

Parameters are dropped

### Possible fixes

Either Fable could be changed to also pass additional arguments or the store could more directly invoke webpack.

### Related information 

* Operating system: Windows 10
* Branch: master

alfonsogarciacaro:
The arguments are passed to Fable CLI. You can redirect some to the subprocess by writing them after `--` as it's customary in other cli tools (like `npm run`).

```shell
dotnet fable yarn-start -- --mode development
```

> BTW, I think `dotnet fable webpack` doesn't work for Webpack 4, as they've changed the main file location. For now, it's better to call Webpack through a package.json script like `start` or `build`.

In any case, I'd recommend setting the Webpack 4 mode in the webpack.config.js file:

```js
module.exports = {
    mode: isProduction ? "production" : "development",
    ...
}
```

The way to pass arguments to the subprocess is also explained in the help text :)

```fsharp
~/d/C/s/Client ❯❯❯ dotnet fable -h
Fable F# to JS compiler (1.3.8)
Usage: dotnet fable [command] [script] [fable arguments] [-- [script arguments]]

Commands:
  -h|--help           Show help
  --version           Print version
  start               Start Fable daemon
  npm-run             Run Fable while an npm script is running
  yarn-run            Run Fable while a yarn script is running
  node-run            Run Fable while a node script is running
  shell-run           Run Fable while a shell script is running
  webpack             Start Fable daemon, invoke Webpack and shut it down
  webpack-dev-server  Run Fable while Webpack development server is running

[...]

PR #314, opened 2018-03-07
Run dev server in webpack development mode

cgravill:
Since V4 it's required to specify the mode
https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a

#313


forki:
are two -- -- correct?

Issue #317, opened 2018-03-20
Upgrade canopy tests to use chrome headless


forki:
AFAIk the phantom is no longer in active development. We should move over to chrome headless.
I'd lvoe to get help for that.

/cc @lefthandedgoat

lefthandedgoat:
ChromeHeadless should work, if it does not let me know and I will try getting the source of this repo and figuring it out.

https://github.com/lefthandedgoat/canopy/blob/master/src/canopy/canopy.fs#L871

forki:
yeah the question is: how would our buld change? I'd really appreciate a PR which shows best pratice ;-)

cgravill:
I have headless Chrome working with Canopy and Expecto on another project. I could draft a PR if that would help?

Rather than check in the drivers I used https://www.nuget.org/packages/Selenium.WebDriver.ChromeDriver/ (and the Gecko and Edge variants) which works well for us. That should avoid the need for the RenameDrivers target as well.

forki:
yes that is exactly what I hoped. getting drivers from nuget or npm.

cgravill:
Looks possible to switch:

```
Starting ChromeDriver 2.37.543627 (63642262d9fb93fb4ab52398be4286d844092a5e) on port 11238
Only local connections are allowed.
[0321/105831.498:ERROR:gpu_process_transport_factory.cc(1019)] Lost UI shared context.
DevTools listening on ws://127.0.0.1:12774/devtools/browser/2db2f1af-53ce-46ae-84a3-472b31dcf62f                        
EXPECTO version
[10:58:34 INF] EXPECTO? Running tests... <Expecto>
[0321/105844.828:INFO:CONSOLE(1)] "Error parsing url: http://localhost:8085/", source: http://localhost:8085/public/bundle.js (1)
[0321/105845.125:INFO:CONSOLE(1)] "Error parsing url: http://localhost:8085/", source: http://localhost:8085/public/bundle.js (1)
[0321/105851.322:INFO:CONSOLE(1)] "Error parsing url: http://localhost:8085/", source: http://localhost:8085/public/bundle.js (1)
[0321/105929.504:INFO:CONSOLE(1)] "Error parsing url: http://localhost:8085/", source: http://localhost:8085/public/bundle.js (1)
[11:00:27 INF] EXPECTO! 4 tests run in 00:01:52.9297318 - 4 passed, 0 ignored, 0 failed, 0 errored. Success! <Expecto>
```
[...]

PR #318, opened 2018-03-22
#317 switch to ChromeHeadless

lefthandedgoat:
I did the best I could testing this.  I could only run via command line, I couldn't get the right combo of stuff installed so that VS would open the project.

Using regular chrome it works and is VERY slow (I stopped it not knowing how many tests there were).  I then switched to ChromeHeadless and it also worked (slowly) but the last test failed.

Concerned about the speed I went to localhost:8085 and there was a noticeable lag as I typed into the box, so I don't think the slowness is related to the changes I made.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -142,20 +142,6 @@ Target "BuildClient" (fun _ ->
 )
 
 // --------------------------------------------------------------------------------------
-// Rename driver for macOS or Linux
-
-Target "RenameDrivers" (fun _ ->
-    if not isWindows then
-        run npmTool "install phantomjs-prebuilt" ""
-    try
-        if isMacOS && not <| File.Exists "test/UITests/bin/Debug/net461/chromedriver" then
-            Fake.FileHelper.Rename "test/UITests/bin/Debug/net461/chromedriver" "test/UITests/bin/Debug/net461/chromedriver_macOS"
-        elif isLinux && not <| File.Exists "test/UITests/bin/Debug/net461/chromedriver" then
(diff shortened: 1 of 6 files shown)


lefthandedgoat:
Looks like the build failed because chrome was not version 64 or above.  I can't tell what version is installed.

You can determine what version is installed, you can find the right chromedriver version here:

https://sites.google.com/a/chromium.org/chromedriver/downloads

Then you can adjust the version of the nuget package to correspond:
 https://www.nuget.org/packages/Selenium.WebDriver.ChromeDriver/

cgravill:
I have AppVeyor working on #319 by downgrading the the driver slightly (Selenium.WebDriver.ChromeDriver = 2.36).

Using the TravisCI Chrome add-in it's attempting to run on TravisCI, however, it looks like it might be an old version of Chrome on there. We might be able to update Chrome that machine to something recent.

cgravill:
@forki I got the tests working on Travis on #319 as well by doing a full install of Chrome. What I did may be a bit outdated compared to https://docs.travis-ci.com/user/gui-and-headless-browsers/ but equally TravisCI's bundled versions of Chrome are now so old as to be problematic. Thoughts?

forki:
can we somehow ask the travis guys to update?

cgravill:
They seem pretty slow at upgrading - latest image is 14.04!

I tied up the approach here so it no longer fakes a display with xvfb and definitely runs fully headless. It’s also using the same approach used to install Yarn. Rather than rely on TravisCI updating this seems like a decent approach.

forki:
that's done now

PR #342, opened 2018-06-18
Elmish.Remoting version

Nhowka:
Communicates with the server through messages via websockets.

Uses a similar `update` function on the server and client to keep a single mindset when designing the application.


alfonsogarciacaro:
Thanks a lot for this @Nhowka! @theimowski Can you please have a look and see if it's possible to add this to the template as an option?

We're starting to have many options available for SAFE stack apps (which is a fantastic thing!) and I wonder what's the best way to promote/explain them in order to avoid confusion. I'm a bit concerned about Fable.Remoting and Elmish.Remoting because it's very likely that some people think of them as the same thing. Could it be possible to combine them somehow (with a clear API that makes the distinction between the different ways of communication)? @Zaid-Ajaj Or maybe add more semantics to the package name (something like `Elmish.Remoting.Websockets` or `Elmish.Remoting.Stateful`)?

Any other ideas? @forki @isaacabraham

Zaid-Ajaj:
@alfonsogarciacaro 

> Could it be possible to combine them somehow (with a clear API that makes the distinction between the different ways of communication)? 

I don't think combining the two libraries is a good idea, they solve client-server communication in very different ways and both have their own mind set. I believe in building small libraries that solve specific problems and doing that well 

> BTW, you can still even both as parts of the same server

@Nhowka 

What do you thing about `Elmish.Messaging`? Sorry if it sounds a bit plain for such an awesome project but when I think of Elmish, I think messages! and the library is all about message passing and reacting to these messages to update state, then you could have `Elmish.Messaging.Suave` and `Elmish.Messaging.Client` etc

Nhowka:
@Zaid-Ajaj I don't hate it, but messages are most of Elmish even when used alone, so it's kinda redundant and give no information about the client-server capabilities. Maybe something flashier could be the way to go. It's a pretty niche project anyway, so it being descriptive and obvious probably won't help that much on the adoption. 🤔

Nhowka:
What is your opinion on `Elmish.Wire`?

alfonsogarciacaro:
Perfect, thanks a lot to you both for the explanations. I love `Elmish.Wire`, it's catchy and perfectly the "tight communication over the network" meaning :clap: :clap:

Zaid-Ajaj:
I like `Elmish.Wire` but I just got one that is more sci-fi 

`Elmish.Fuse - the ultimate client-server elmish fusion`  hahah (I like this one for real :smile:) 

the rest is up to you @Nhowka

Nhowka:
I like the full title, but the other meaning of `Fuse` is about a wire that works by breaking. 😅

Other options I thought were `Elmish.Connect` and `Elmish.Link`. I'm liking link more as it is as short and it feels right. It links the server and client with the same architecture on both sides. I'll open an issue for it, maybe Link is overused for stuff that I'm not aware...

isaacabraham:
Elmish.Fusion? Elmish.Bridge?

I think wire is pretty good though!

Nhowka:
Issue is open! https://github.com/Nhowka/Elmish.Remoting/issues/1

I'll post about that on twitter/Slack too.

theimowski:
I added https://github.com/SAFE-Stack/SAFE-template/issues/107 to track discussion on adding that to the template

alfonsogarciacaro:
@Nhowka Are you going to update the branch with the new name? Maybe you can also send a PR to master's README to add a link to this branch.

Nhowka:
@alfonsogarciacaro Yes, I'll try to do that soon.

Nhowka:
Updated on #351. Not sure what to write on the README yet, but as that is in another branch, I'll do it another time.

PR #349, opened 2018-07-17
Update to Fable 2 alpha

MangelMaxime:
The project is now compiling and working but several things are needed before considering Fable 2 ready for booktstore.

- [x] Release a new version of Fable.PowerPack that is using Thoth.Json
    - This is needed in order to fix `Fetch.fetchAs<'T>` which failed silently as it's now just using JavaScript `parse`
- [x] Create `Thoth.Json.Net.Giraffe` to host the `ThothSerializer` type
- [x] Clean the code
  - Remove logs
  - Remove dead code


MangelMaxime:
This is ready to be review and for others to explore how to structure the app.

I manually "unwrap" the fetch result because I wasn't able to add a new file for the client.

Things like:

```fs
        let! res = Fetch.fetch url props
        let! txt = res.text()
        return Decode.Auto.unsafeFromString<WishList> txt
```

MangelMaxime:
This is ready to be review and for others to explore how to structure the app.

I manually "unwrap" the fetch result because I wasn't able to add a new file for the client.

For things like:

```fs
        let! res = Fetch.fetch url props
        let! txt = res.text()
        return Decode.Auto.unsafeFromString<WishList> txt
```

it should be possible to do something like:

```fs
        return! Fetch.fetchAs url Decoder.wishList props
```

I wanted to create a file `Decoder.fs` specific for the client which would have contain things like:

```fs
module Decoder 

open Thoth.Json

let wishList = Decoder.Auto.generateDecoder<WishList>()
```

So the decoder information is stored only once in the `bundle.js` to reduce the JavaScript size. However, I wasn't able to do that. When adding a new entry to `Client.props` the file was gaining access to the packages. If someone want to try it :)

I am not used to the SAFE stack folder/file architecture.

MangelMaxime:
I will now rebase this PR with master now that Fable 2 has been released.

forki:
@MangelMaxime can you please resolve conflicts?

package.json
```diff
@@ -14,9 +14,9 @@
     "babel-plugin-transform-runtime": "6.23.0",
     "babel-preset-env": "1.7.0",
     "concurrently": "3.6.0",
-    "fable-loader": "1.1.6",
-    "fable-utils": "1.0.6",
-    "webpack": "4.15.1",
+    "fable-loader": "^2.0.0-alpha-003",
+    "uglifyjs-webpack-plugin": "^1.2.7",
```
forki: why do we need uglify?

paket.dependencies
```diff
@@ -11,18 +10,22 @@ nuget Microsoft.AspNetCore.WebSockets
 nuget Microsoft.AspNetCore.WebSockets.Server
 nuget System.Net.NetworkInformation
 nuget jose-jwt
+nuget Thoth.Json.Net prerelease
+nuget Thoth.Json.Giraffe
 
-nuget Fable.Core
-nuget Fable.React
+nuget Fable.Core prerelease
```
forki: don't need prerelease any more

package.json
MangelMaxime: Seems like it is causing some problem with inline method.

@alfonsogarciacaro Will be able to tell us if it's still the case or not.

MangelMaxime: https://github.com/SAFE-Stack/SAFE-BookStore/pull/349/files#diff-19e2e2ba7026ac670b56c4cb0781b7fbR79

paket.dependencies
MangelMaxime: Yes, working on it right now.

MangelMaxime:
@forki I have a runtime error when running the server and I don't know how to fix it.

```shell
 Method not found: 'System.String IJsonSerializer.Serialize(System.Object)'.
/Users/maximemangel/Workspaces/Github/SAFE-Stack/SAFE-BookStore/test/ServerTests/ServerTests.fsproj(16,5): error MSB3073: La commande "dotnet run" s'est arrêtée avec le code 1.
watch : Exited with error code 1
watch : Waiting for a file to change before restarting dotnet...
```

If I run `dotnet build` against the server it's working.

Things left to do:
- Fix the server
- If we want to minimize a bit more the bundle size add a `Decoder.fs` file specific to the `Client.fsproj`

    The file should looks like:

    ```fs
    module Decoder 

    open Thoth.Json

    let wishList = Decoder.Auto.generateDecoder<WishList>()
    ```

    More information [here](https://mangelmaxime.github.io/Thoth/json/v2/decode.html#size-optimization)

MangelMaxime:
Ok @Nhowka helped me with the server error. I think we now know the source of it.

I have a fix in progress and will test/apply it tomorrow.

MangelMaxime:
@forki The PR is ready for review.

Travis CI is failing but I don't think this is something related to my changes because it's the setup of the container that is failing.

src/Client/webpack.config.js
```diff
@@ -1,76 +1,116 @@
...
+    },
+    mode: isProduction ? "production" : "development",
+    devtool: isProduction ? undefined : "source-map",
+    optimization: {
+        // Split the code coming from npm packages into a different file.
+        // 3rd party dependencies change less often, let the browser cache them.
+        splitChunks: {
+            cacheGroups: {
+                commons: {
+                    test: /node_modules/,
```
0x53A: What do you think about changing this to ``/node_modules|\.nuget/``?

I did a quick check with my own app:

|    file  | /node_modules/ | ``/node_modules\|\.nuget/`` |
| ---------- | ------ | ------ |
| main.js    | 157 kb | 30 kb  |
| vendors.js | 236 kb | 382 kb |
| total | 393 kb | 412 kb |

So the total size is ~5% greater, but the main bundle is much smaller.

src/Client/webpack.config.js
```diff
@@ -1,76 +1,116 @@
...
-        target: 'http://localhost:' + port,
-        changeOrigin: true
-      }
+    entry: CONFIG.fsharpEntry,
+    // NOTE we add a hash to the output file name in production
+    // to prevent browser caching if code changes
+    output: {
+        path: CONFIG.outputDir,
+        publicPath: "/public",
+        filename: '[name].js'
```
marcpiechura: Shouldn’t this be either [name][hash].js for production or [name].js for development builds ?

src/Client/webpack.config.js
MangelMaxime: I am surprised that `.nuget` is the rule to apply here... Because in theory Fable 2 is using the `.fable` created locally.

Splitting between `vendors.js` and `main.js` is to optimize the browser caching. If you include `nuget` (aka Fable libraries in it) then if you use a new Fable feature or a new feature from a library then the `vendors.js` will change and will be loading from the server and not the cache.

Also, in general in the different projects, I use the hashed names but here because we are supporting SSR then I am not using the hashed names. We could support hashed names if we generate the `main.[hash].js` and `vendors.[hash].js` first and then update a file on the server to register the generated names.

To answer more directly your question, I am not against including more things in the `vendors.js` I think it's up to the people maintaining the SAFE stack to step in and discuss/decide.

src/Client/webpack.config.js
MangelMaxime: The problem is about supporting SSR.

> Also, in general in the different projects, I use the hashed names but here because we are supporting SSR then I am not using the hashed names. We could support hashed names if we generate the `main.[hash].js` and `vendors.[hash].js` first and then update a file on the server to register the generated names.

forki:
awesome work!

src/Client/App.fs
```diff
@@ -12,9 +12,7 @@ open Elmish.HMR
 open Client.Shared
 open Client.Pages
 open ServerCode.Domain
-
-JsInterop.importSideEffects "whatwg-fetch"
```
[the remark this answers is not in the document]
MangelMaxime: Directly added via webpack: https://github.com/SAFE-Stack/SAFE-BookStore/pull/349/files#diff-19e2e2ba7026ac670b56c4cb0781b7fbR10

src/Client/webpack.config.js
0x53A: I'm still using Fable 1.x, so with fable 2.x it would probably be ``.fable`` as you say.


``vendors.js`` is still tree-shaken, right? So if I don't use a feature from a node lib, it would also get added / removed?

There is a manifest plugin for webpack - it will create a json file which maps from unhashed to hashed: https://github.com/webdeveric/webpack-assets-manifest

On the SSR side, this json file would need to be read.

MangelMaxime: > `vendors.js` is still tree-shaken, right?

It should depending on the library. Also, we kind of consider that the `npm` usage change less often. Compare to Fable code but we can be wrong. Also, in general cache mechanism should be optimize depending the project setup/needs. It's hard to find a perfect rule for every one because each way have it's pros and cons. Also, I am not an expert of this things so can be wrong too :)

Indeed the maniafest plugin need to be added when using hashed name so the same content generate the same hash.

src/Client/App.fs
```diff
@@ -146,6 +148,7 @@ Program.mkProgram init update view
 #endif
 |> withReact "elmish-app"
 #if DEBUG
-|> Program.withDebugger
+// The debugger isn't support yet for Fable 2
```
forki: @MangelMaxime @alfonsogarciacaro what about this?

MangelMaxime: @forki It's now working with Fable 2. 

You can update to latest version and use it like before.

src/Server/Storage/FileSystem.fs
```diff
@@ -12,10 +13,10 @@ let getWishListFromDB userName =
...
-        |> FableJson.ofJson<WishList>
+        |> Decode.Auto.unsafeFromString<WishList>
 
 let saveWishListToDB wishList =
     let fi = FileInfo(getJSONFileName wishList.UserName)
     if not fi.Directory.Exists then
         fi.Directory.Create()
-    File.WriteAllText(fi.FullName, FableJson.toJson wishList)
\ No newline at end of file
+    File.WriteAllText(fi.FullName, Encode.Auto.toString(4, wishList))
```
forki: @MangelMaxime  what's the 4 supposed to mean here?

MangelMaxime: It's the number of spaces used for indendation.

In general, I set it to `4` in DEBUG mode and `0` for production usage.

forki:
I see. Thx

Am Mi., 7. Nov. 2018, 13:01 hat Maxime Mangel <notifications@github.com>
geschrieben:

Issue #359, opened 2018-08-23
Add typescript source component


0x53A:
I found that it is actually pretty easy to integrate typescript source components into the build pipeline with webpack.

What do you think about adding an example of this?

My idea would be to create one react component in .tsx, then use it from f#.

-------

I think that showing great interop is one of the more important points when advertising fable. F# and Elmish are an awesome combination by themselves, but at first glance it maybe feels like a bit of a walled garden.

PR #361, opened 2018-08-23
[WIP] Add typescript source component

0x53A:
this shows the minimum configuration to add a typescript component into the webpack build pipeline

image

I'd define one react component in typescript and reuse it in F#.

What do you think? Would it make sense to integrate this into the BookStore, or should I just write a blog post? ;D

### Todo

[ ] Actually add a React component

### Issues

[ ] I can't reference one tsx from another, when I do ``import X from './test2'``, where there is a file `test2.tsx` next to it, it fails with 

```
Module not found: Error: Can't resolve './test2'
```

closes https://github.com/SAFE-Stack/SAFE-BookStore/issues/359

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "remotedev": "0.2.7"
   },
   "devDependencies": {
+    "awesome-typescript-loader": "^5.2.0",
     "babel-core": "6.26.3",
     "babel-loader": "7.1.5",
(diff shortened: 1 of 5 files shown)


forki:
Mhm dunno. I personally think blog post would be better

0x53A:
Ok, makes sense to focus this project on the fable part

PR #363, opened 2018-08-29
Add serverside rendering for /login

0x53A:


0x53A:
SSR was already enabled for ``/``, so I don't think so?

I did ``build runssr``, and it seemed to work ok.

Anything special you want me to check?

Of course the SSR is only active if I directly navigate to ``/login``, if I first navigate to ``/``, then click on login, it is rendered client-side (as it should be).

------

I wanted to make it clear that SSR can - in theory - be used for any page, just the home page is not that much.

forki:
https://github.com/SAFE-Stack/SAFE-BookStore/blob/master/src/Client/App.fs#L58

That's where the hydrate part is kicking in. For login it does not make any difference since that page doesn't do any further web requests. But in general case we want to hydrate the server side calculated model into the runtime js model. 

An example could be a public wishlist. You would render it on serverless side and additionally you would hydrate the array of books into the model. So you would save that additional round trip.

forki:
And ftr SSR with hydrate is awesome! That's all.

0x53A:
Ah, didn't know that ;D

I added some comments, what do you think?

0x53A:
Do you have any public examples for a complicated hydrate @forki ?

forki:
Nothing public. Maybe @zaaack has something. And I think we should still hydrate all SSR pages. This will ensure compiler complains if we change model later to something more complicated

0x53A:
> And I think we should still hydrate all SSR pages. This will ensure compiler complains if we change model later to something more complicated

Like this?

```fsharp
let hydrateModel (json:string) : Model * Cmd<_> =
    // The page was rendered server-side and now react client-side kicks in.
    // If needed, the model could be fixed up here.
    // In this case we just deserialize the model from the json and don't need to to anything special.
    let model : Model = ofJson json
    match model.PageModel with
    | HomePageModel -> model, Cmd.none
    | LoginModel _ -> model, Cmd.none
| WishListModel _ -> model, Cmd.none
```

forki:
Yep

0x53A:
ah, I didn't even see the ``Some Page.Home``.

forki:
Uh something is red?

0x53A:
Net SDK download failure 

https://travis-ci.org/SAFE-Stack/SAFE-BookStore/builds/422204639#L2343

forki:
💯

PR #367, opened 2018-10-03
Use manual encoders/decoders

MangelMaxime:
Asked by @forki on Gitter

> I wonder if you could send a pr to bookstore that actually uses Toth decoders and removes the generic ones

Possible improvements:
- Use an alias for `Encode.toString`. Like that in dev mode we have readible JSON and in production they are optimized
    ```fs
    module Encode =
        let inline toString (value: Value) : string =
        #if DEBUG
        Encode.toString 4 value
        #else
        Encode.toString value
        #endif
    ```

Because we can't set compiler directive for intellisense at least in Ionide we need to reference `Thoth.Json.Net` for some special place in the client. Perhaps, there is another way to do it but I don't know it. If you want to try something here is the place where I needed this:
- [Decode model from SSR](https://github.com/SAFE-Stack/SAFE-BookStore/compare/master...MangelMaxime:feature/manual_encoders_decoders?expand=1#diff-73aaaba26167c7b73ef74ad945c93610R70)
- [Manually access the localStorage](https://github.com/SAFE-Stack/SAFE-BookStore/compare/master...MangelMaxime:feature/manual_encoders_decoders?expand=1#diff-73aaaba26167c7b73ef74ad945c93610R51): needed here because `BrowserLocalStorage.load` is a Fable library.

About this limitation some people in the past mentioned that we could perhaps create a `Common` library to use the same "type references". [for example](https://github.com/MangelMaxime/Thoth/issues/45). However, I don't know how to write such a common layer.

I decided to go fully with using manual decoders but a solution to avoid this intellisense problem could be to use `AutoDecoders` only for the SSR part. And use manual decoders for the rest of the Project. It can be a solution and don't impact too much the bundle size because we would need to generation the reflection data only at one place in the Client.


isaacabraham:
LGTM, if this is the way that we want to bookstore to go :-) The only question is that the indentation for some |> looks a bit atypical to me - but that could just be the github diff.

MangelMaxime:
@isaacabraham Feel free to correct it (you have the right to commit to my branch as a maintainer) or make a review of with annotations I will update accordingly.

MangelMaxime:
And yes for F# to F# JSON deserialization/serialization it's probably not needed to use manual encoders/decoders as long as you share the same Type definition between both projects.

However, if you are interacting with an external JSON provider then using manual encoders/decoders is the way to go so you don't need to do a 1 to 1 mapping between the JSON and F# types. 

For example, at work, I decode the JSON data directly into a structure that is meaningful for my `Model` which isn't following the JSON structure. I flatten the data structure or add "state information" like (isLoading, isEditing, etc.) which isn't part of the JSON.

src/Server/WishList.fs
```diff
@@ -7,13 +7,16 @@ open Giraffe
...
 
 /// Handle the GET on /api/wishlist
 let getWishList (getWishListFromDB : string -> Task<WishList>) (token : UserRights) : HttpHandler =
      fun (next : HttpFunc) (ctx : HttpContext) ->
         task {
             let! wishList = getWishListFromDB token.UserName
-            return! ctx.WriteJsonAsync wishList
+            let json = WishList.Encoder wishList
+                        |> Encode.toString 0
+            return! ctx.WriteStringAsync json
```
forki: this doesn't set the HTTP headers correctly, right!?

MangelMaxime: Probably no, I don't know Giraffe.

I guess adding an extension method like 

```fs
ctx.WriteJsonAsync(jsonValue : Encode.Value)
``` 

could be something to do in the future for SAFE apps.

You could also pass it the `value` and the `encoder` to apply:

```fs
ctx.WriteJsonAsync(value : 'T, Encode.Encoder<'T>)
```

I find the second option easier to read because people don't need to apply the `encoder` themself.

forki:
@Krzysztof-Cieslak could you please look at this an saturnize it?! that would be really great

src/Server/WishList.fs
alfonsogarciacaro: See [this version](https://github.com/MangelMaxime/Thoth/issues/80#issuecomment-441707345). This sets the content type header and the status code. And should be more performant as well :)

alfonsogarciacaro: See [this version](https://github.com/MangelMaxime/Thoth/issues/80#issuecomment-441707345). This sets the content-type header and the status code. And should be also more performant :)

isaacabraham: We should just be using the `json` helper in Giraffe, having plugged in the appropriate Thoth helper at the start. The SAFE Template does this so there's a pattern that can be followed I think.

alfonsogarciacaro: @isaacabraham The problem, [as noted here](https://github.com/MangelMaxime/Thoth/issues/80#issuecomment-441591039), is the Giraffe.ThothSerializer only works with auto encoders. It can conflict with manual coders if they're used.

isaacabraham: Doh, sorry, I missed that.

PR #372, opened 2018-11-06
with debugger

forki:
/cc @MangelMaxime

### DIFF
--- a/paket.lock
+++ b/paket.lock
@@ -1,22 +1,14 @@
 STORAGE: NONE
 NUGET
   remote: https://www.nuget.org/api/v2
-    dotnet-fable (1.3.19) - clitool: true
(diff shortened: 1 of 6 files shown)


MangelMaxime:
Yes should be working

forki:
oh it's red. any idea what's broken with webpack?

paket.lock
```diff
@@ -1,17 +1,25 @@
 STORAGE: NONE
 NUGET
   remote: https://www.nuget.org/api/v2
-    dotnet-fable (2.0.2) - clitool: true
+    dotnet-fable (1.3.19) - clitool: true
```
MangelMaxime: @forki This version is incorrect you are back to Fable 1

forki: urgs

PR #375, opened 2018-11-16
Update webpack.config.js

0x53A:
pass babel configuration to fable-loader

see https://github.com/SAFE-Stack/SAFE-template/issues/188


0x53A:
interesting error
grafik

forki:
Is this something we need to do in downstream projects?

0x53A:
Afaik yes.

PR #385, opened 2019-03-14
Tiny fix

tforkmann:


tforkmann:
Appveyor fails because it can't download FAKE 4

PR #388, opened 2019-04-12
WIP: Update Fable 2.2 & Fable.Core v3

MangelMaxime:
@forki @alfonsogarciacaro I am trying to update the repo to use Fable 2.2 and Fable.Core v3.

For now, I have just updated the client side to Fable.Core v3 however I am stuck.

In order, to make paket resolve the packages I needed to add `framework: netstandard2.0` to `paket.dependencies` otherwise `Microsoft.NET.Sdk.Functions` complains about a conflict for the resolution of dependencies for `.Net Framework`.

But if I try to build the server, it's failing and find no dependency. I guess it's because of `framework: netstandard2.0` while the server is a `netcoreapp2.2` so I tried `framework: netcoreapp2.2,  netstandard2.0` without success.

Would you please be able to guide me on how to solve the situation or contribute to this PR?

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -17,14 +17,35 @@
     <PaketExeImage Condition=" '$(PaketBootstrapperStyle)' == 'proj' ">native</PaketExeImage>
     <MonoPath Condition="'$(MonoPath)' == '' And Exists('/Library/Frameworks/Mono.framework/Commands/mono')">/Library/Frameworks/Mono.framework/Commands/mono</MonoPath>
     <MonoPath Condition="'$(MonoPath)' == ''">mono</MonoPath>
-    <!-- Paket command -->
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND Exists('$(PaketToolsPath)paket')">$(PaketToolsPath)paket</PaketExePath>
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND Exists('$(PaketRootPath)paket.exe')">$(PaketRootPath)paket.exe</PaketExePath>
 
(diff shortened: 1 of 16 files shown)


forki:
Will take a look at this on Monday

MangelMaxime:
For info, I already had to lower `Newtonsoft` required version by `Thoth.Json.Net` in order to make it resolve the package. I opened an issue for `Giraffe` too https://github.com/giraffe-fsharp/Giraffe/issues/350

alfonsogarciacaro:
We should create a different group for Fable/Client dependencies so you can add `framework: netstandard2.0` without affecting the Main/Server group. This is what I do in my full stack projects and it's working nicely :)

MangelMaxime:
This means using differents packages versions on the server and the client I am not sure if that's really a good idea^^.

Also, I wonder why now it's needed while before it was not. And if that an easy fix to go back the previous was I think it's better.

alfonsogarciacaro:
Different versions? You mean Thoth.Json? That's the only one package "shared" between frontend and backend, right? You can pin the version or create a specific group for it (not sure if FSharp.Core can give issues in that case).

forki:
Yes introducing different groups is good. In my projects I have one for
azure functions, one for client, one for server and another one for build

Alfonso Garcia-Caro <notifications@github.com> schrieb am Fr., 12. Apr.
2019, 18:16:

forki:
let's continue with: https://github.com/SAFE-Stack/SAFE-BookStore/pull/389

MangelMaxime:
@alfonsogarciacaro Well there is also `Fable.React` which is shared. But if you guys thin that's fine then let's go this way :)

MangelMaxime:
Ok I see we use a group to separate the Azure runtime so my previous comment is invalid :)

PR #390, opened 2019-04-15
Update Fable.React

alfonsogarciacaro:
This updates to latest Fable.React that makes it easier to create function components. I will publish the stable version soon, so you may want to wait until that before merging the PR.


src/Client/pages/WishList.fs
```diff
@@ -147,7 +147,8 @@ let bookComponent { book = book; removeBook = removeBook } =
         td [] [ buttonLink "" removeBook [ str "Remove" ] ]
     ]
 
-let inline BookComponent props = (ofFunction bookComponent) props []
+let BookComponent =
+    FunctionComponent.Of bookComponent
```
forki: nice

PR #431, opened 2023-02-27
Upgrade everything!

jwosty:
* Upgrade to .NET 6
* Upgrade to FAKE 5
* Upgrade to Paket 7
* Upgrade to Fable 3
* Upgrade to Azure SDK 12
* Upgrade pretty much everything else too

Some help verifying that the app still works as expected would be really great. I'm not familiar with the Azure SDKs, so I made some educated guesses when dealing with the breaking changes in the latest versions.


jwosty:
@forki any idea how to fix this latest build failure? It seems to not be finding the Chrome binary when running the UI tests in the CI, I'm not sure if this is just some kind of version mismatch or what. This is not my area of expertise :)

cgravill:
> @forki any idea how to fix this latest build failure? It seems to not be finding the Chrome binary when running the UI tests in the CI, I'm not sure if this is just some kind of version mismatch or what. This is not my area of expertise :)

@jwosty yes it's very likely to be a version mismatch. The change of AppVeyor image means it now has a much newer version of chrome: https://www.appveyor.com/docs/windows-images-software/#:~:text=Chrome%20110.0.5481.178

In
https://github.com/jwosty/SAFE-BookStore/blob/e938e77f24f22bcd03836e9926702b90f3f143f9/paket.dependencies#L44-L50

I suggest changing Selenium.WebDriver.ChromeDriver version pinning to match i.e. something like:
https://www.nuget.org/packages/Selenium.WebDriver.ChromeDriver/110.0.5481.7700

There's also a minor bump to Canopy that might improve odds of it working too.

(great to see this update happening)

jwosty:
@cgravill Hmm, I was afraid that would be the case... I had tried upgrading Canopy and Selenium to the latest versions of both (see [aa9801b](https://github.com/SAFE-Stack/SAFE-BookStore/pull/431/commits/aa9801beb10365b69ce9e35508516bf44b9f19d6)), but was getting a different problem (see https://ci.appveyor.com/project/isaacabraham/safe-bookstore/builds/46344971#L646). Some debugging uncovered that Canopy was blowing up ([on this line](https://github.com/lefthandedgoat/canopy/blob/5cc982f7eed684a407ee67b492aecf9a50b8c4ea/src/canopy/canopy.parallell.functions.fs#L949)) when trying to execute the Javascript snippet `return window.innerWidth`, which was returning a dictionary with a key `message` and a value:

```
no such session
  (Driver info: chromedriver=2.36.540469 (1881fd7f8641508feb5166b7cae561d87723cfa8),platform=Mac OS X 10.16.0 x86_64)]
```

Which is why I had attempted to downgrade to previously working package versions. Looks like we really will have to use latest Chrome after all, though. I'm sure it's due to breaking changes in the Chromium API. Any ideas on how to fix this one?

jwosty:
Ah, I see that there was a mistake in paket.dependencies causing the confusion. Now we're just onto a different problem!

cgravill:
Brilliant looks like it's running now.

I had a quick look at the current error and it's failing to find elements for the tests - probably the quickest way to resolve is to manually inspect the html running locally to see if things have been moved around by version changes. If it's not obvious I could try to take a look sometime.

Issue #448, opened 2024-02-23
Use React Components for inner pages


isaacabraham:
The page uses the basic Elmish pattern for outer / inner pages. This works and is simple but quite chatty and hard to read.

We should move to making inner pages their own React components.

PR #451, opened 2024-03-01
Use elmish

Larocceau:
in response to #448:

* Use UseElmish hook for pages instead of passing messages on from index


martinbryant (approved):
Reviewed with @Larocceau and issues resolved

